PR #10, opened 2017-12-30
Add to_slice and to_vec functions

uzytkownik:
I needed to convert to/from mpz_t. While converting into is relativly simple due to `from_slice` function the invert requires allocation of array with base 2<sup>8</sup> making it less efficient (requiring 2 copies and 2 convertions from/to 8 bit instead of 1 or 0).

Those functions would allow inspecting representation of function without need of allocating and creating additional copies.


uzytkownik:
Temporary closing to change branch.

PR #12, opened 2018-01-03
Add to_slice and to_vec functions

uzytkownik:
I needed to convert to/from mpz_t. While converting into is relativly simple due to from_slice function the invert requires allocation of array with base 28 making it less efficient (requiring 2 copies and 2 convertions from/to 8 bit instead of 1 or 0). Those functions would allow inspecting representation of function without need of allocating and creating additional copies.

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -1540,6 +1540,42 @@ impl BigInt {
         BigUint::from_radix_le(buf, radix).map(|u| BigInt::from_biguint(sign, u))
     }
 
+    /// Returns the representation of the `BigInt` in little-endian base 2<sup>32</sup>.
+    ///
+    /// # Examples
+    ///
+    ///
+    /// ```
+    /// use num_bigint::{Sign, BigInt, BigDigit};
+    ///
+    /// let sign = Sign::Plus;
+    /// let slice: &[BigDigit] = &[1, 2, 3, 4];
+    /// let i = BigInt::from_slice(sign, slice);
+    /// assert_eq!(i.to_slice(), (sign, slice));
+    /// ```
+    #[inline]
+    pub fn to_slice(&self) -> (Sign, &[BigDigit]) {
+        (self.sign, self.data.to_slice())
+    }
+
+    /// Returns the representation of the `BigInt` in little-endian base 2<sup>32</sup>.
+    ///
+    /// # Examples
+    ///
+    ///
+    /// ```
+    /// use num_bigint::{Sign, BigInt};
+    ///
+    /// let sign = Sign::Plus;
+    /// let vec = vec![1, 2, 3, 4];
+    /// let i = BigInt::from_slice(sign, &vec);
+    /// assert_eq!(i.to_vec(), (sign, vec));
+    /// ```
+    #[inline]
+    pub fn to_vec(self) -> (Sign, Vec<BigDigit>) {
+        (self.sign, self.data.to_vec())
+    }
+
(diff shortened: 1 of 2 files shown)


cuviper:
I hesitate to expose the internal representation so much, for the borrowed slice especially.

How would you feel about some kind of iterator instead? We could make it return bytes or even larger units, without having to allocate the full conversion.

uzytkownik:
That would not solve the problem fully due to [`mpz_export`](https://docs.rs/gmp-mpfr-sys/1.0.8/gmp_mpfr_sys/gmp/fn.mpz_export.html)/[mpz_import](https://docs.rs/gmp-mpfr-sys/1.0.8/gmp_mpfr_sys/gmp/fn.mpz_export.html) functions. GMP documents internals only partially (namely - there are some compile-time constants) so I guess it should work assuming some support for [nails](https://docs.rs/gmp-mpfr-sys/1.0.8/gmp_mpfr_sys/gmp/constant.NAIL_MASK.html),

cuviper:
Perhaps we could frame this in a way that doesn't commit to a particular representation?  Like:

```rust
impl BigUint {
    // Returns `(count, order, size, endian, nails, ptr)` in the style of `mpz_import`
    fn export_raw<F, T>(&self, f: F) -> (usize, c_int, usize, c_int, usize, *const c_void);
}
```

Obviously that's tailored directly for GMP, but I'm not sure how to generalize this idea.

tspiteri:
If what is required is conversion to `mpz_t`, an iterator could work, considering that GMP does document its internal representation. As a note, I don't think nails need to be supported, considering that if I'm not mistaken, compiling GMP with nails has been broken since late in the 4.* series (that's ancient). I hacked up something assuming that `BigUint` has some `digits()` method that gives an iterator with the digits of our requested size (`u32` or `u64`), least significant first. I did not test it, this is just to give an idea of how to go about it:

```rust
use gmp_mpfr_sys::gmp;
use std::slice;
fn copy_to_gmp(src: &BigUint, dst: &mut gmp::mpz_t) {
    assert_eq!(gmp::NAIL_BITS, 0, "nails not supported");
    let bits = src.bits();
    if bits == 0 {
        // Since we're going to poke into the mpz_t internals, might
        // as well do it here too.
        dst.size = 0;
        return;
    }
    let limb_bits = gmp::LIMB_BITS as usize;
    let limb_count = (bits + limb_bits - 1) / limb_bits;
    let limbs = unsafe {
        gmp::_mpz_realloc(mpz_t, limb_count);
        dst.size = limb_count;
        slice::from_raw_parts_mut(dst.d, limb_count)
    };
    // limb_t is either u32 or u64
    let mut digits = src.digits::<gmp::limb_t>();
    // Since the slice was created explicitly with size limb_count,
    // the compiler does not introduce runtime checks for limbs[i].
    for i in 0..limb_count {
        limbs[i] = digits.next().expect("not enough digits");
    }
    assert!(digits.next().is_none(), "too many digits");
}
```

dignifiedquire:
Another data point: it would be great to have the limbs exposed as an iterator

Speedy37:
I've the same issue, I need to access the internal &[u32] slice.
There is a `from_slice`, so a `to_slice` function returning a Vec<u32> would be nice and doesn't require to commit to a predefined internal representation.

For performance, it would be nice to have an iterable version of it or a non documented version (like `as_slice`).

expenses:
I'd like either a function that returned the underlying slice or an iterator too.

PR #20, opened 2018-02-09
Note a few alternative crates in README.md

cuviper:
This presents a table of some common alternatives to this crate which may have better performance.  I'm not including any specific performance comparisons, since that will surely vary on different workloads, and will change as each crate evolves.  I want to keep this impartial, listing only simple objective facts about each.  Corrections welcome!

cc crate owners: @Aatch, @doomrobo, @fizyk20, @tspiteri


cuviper:
I'm not sure if `gmp-mpfr-sys` really belongs on this list, given that `rug` appears to be its only dependent.  Maybe I should remove that and avoid the distraction of safe/unsafe.

Issue #4, opened 2017-12-19
Implement std::iter::{Sum,Product} 


cuviper:
(copying from rust-num/num#233, here for bigint types only)

The [Sum](https://doc.rust-lang.org/std/iter/trait.Sum.html) and [Product](https://doc.rust-lang.org/std/iter/trait.Product.html) traits are stable in 1.12.0, so we could implement them for our types.

(But we consider it a breaking change to increase the required rustc.)

cuviper:
This was started in rust-num/num#348.

cuviper:
Closed by #22.

PR #22, opened 2018-02-22
Implementation of std::iter::{Product, Sum}

dodomorandi:
This PR is relative to the [issue 4](https://github.com/rust-num/num-bigint/issues/4). This is a
very simple implementation, based on two macros. Maybe a more elegant
solution is possible for these, feel free to suggest.

Two tests for both `BigInt` and `BigUInt` have been added, one relative
to the `Sum`, the other to the `Product`.

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -6,6 +6,7 @@ use std::cmp::Ordering::{self, Less, Greater, Equal};
 use std::{i64, u64};
 #[allow(unused_imports)]
 use std::ascii::AsciiExt;
+use std::iter::{Product, Sum};
 
 #[cfg(feature = "serde")]
 use serde;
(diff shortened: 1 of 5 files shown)


cuviper:
I think adding such a generic impl might itself be a breaking change, as someone could have their own `impl Sum<Foo> for BigUint` already -- but that's ok if we're rolling this into 0.2.

dodomorandi:
I really liked your suggestion, so I decided to make the two implementations generic. Two additional tests have been added to check against _generic_ `Sum` and `Product` (in this case from a vector of `i32` or `u32`, depending from the case).

I want to ask you a thing about conditional compilation: is there a way to check against _std_ features using `cfg` (`iter_arith_traits` in this case)? I am quite new with Rust, and I never needed to perform this kind of check... until now. I am not able to find anything useful, but maybe it is just me. Do you have any idea?

cuviper:
> is there a way to check against std features

Sadly, no.  So many things would be easier if there were.

cuviper:
I rebased now that 0.2 has started, and fixed on issue with rust 1.15.

Thanks for implementing this!

bors r+

PR #28, opened 2018-02-27
Implement is_one for BigUint and BigInt.

clarcharr:
Won't work until a newer version of `num-traits` is pushed, but I figured I'd offer these now.


cuviper:
This should be good with a bump to num-traits 0.2.1.

PR #30, opened 2018-03-01
reduce copying when forwarding ref-ref to val-ref

tspiteri:
When ref-ref is forwarded to val-ref, currently the larger of the operands is cloned. This is good in one way as the target has the required capacity and will not need reallocation. However, it involves more copying before the actual operation starts.

With this commit, the clone will have the capacity of the larger operand, but only the data for the smaller operand is copied, saving some copying if the operand sizes are very different.

For addition I don't think this commit would result in much gain, because once the smaller operand is used up, the addition stops, so the copying is actually used. The commit shouldn't be a loss either, and maybe it could give a small gain for very large sizes, as the lower digits could maybe still be in the cache for the addition before moving on to copying the higher digits.

For some bitwise operations this could give a much larger gain. For example when performing a bitwise and of a large number with a small number, the higher digits would be copied uselessly without this commit.

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -27,7 +27,7 @@ use super::ParseBigIntError;
 use super::big_digit::{BigDigit, DoubleBigDigit};
 use biguint;
 use biguint::to_str_radix_reversed;
-use biguint::BigUint;
+use biguint::{BigUint, CloneWithCapacity};
 
 use UsizePromotion;
 use IsizePromotion;
@@ -64,6 +64,21 @@ impl Mul<Sign> for Sign {
             (Plus, Plus) | (Minus, Minus) => Plus,
             (Plus, Minus) | (Minus, Plus) => Minus,
         }
+    }
+}
+
+impl CloneWithCapacity for BigInt {
+    #[inline]
+    fn capacity(&self) -> usize {
+        self.data.capacity()
+    }
+    #[inline]
+    fn len(&self) -> usize {
+        self.data.len()
+    }
+    #[inline]
+    fn clone_with_capacity(&self, capacity: usize) -> BigInt {
+        BigInt::from_biguint(self.sign, self.data.clone_with_capacity(capacity))
     }
 }
 
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -1689,6 +1689,29 @@ pub fn trailing_zeros(u: &BigUint) -> Option<usize> {
 impl_sum_iter_type!(BigUint);
 impl_product_iter_type!(BigUint);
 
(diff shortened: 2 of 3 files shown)


tspiteri:
I redid the commit in such a way that it is compatible with `BigInt` as well, which could be helpful eventually.

tspiteri:
So I tried to benchmark this and did a small test: benchmarking an operation with `lhs = fib(10000)` and `rhs = &lhs * &lhs`. The logic and `&lhs & &rhs` was indeed a little better, about 100 ns vs 115 ns, but the addition `&lhs + &rh2` regressed a lot, as in going from 320 ns to 580 ns.

The problem was that the implementation of short_val + long_ref is extending the vector of short_val with zeros before adding, so that `__add2` gets a mutable slice at least as long as the immutable one. I changed the `AddAssign` implementation like this:
```diff
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -433,12 +433,15 @@ impl<'a> Add<&'a BigUint> for BigUint {
 impl<'a> AddAssign<&'a BigUint> for BigUint {
     #[inline]
     fn add_assign(&mut self, other: &BigUint) {
-        if self.data.len() < other.data.len() {
-            let extra = other.data.len() - self.data.len();
-            self.data.extend(repeat(0).take(extra));
-        }
-
-        let carry = __add2(&mut self.data[..], &other.data[..]);
+        let self_len = self.data.len();
+        let other_len = other.data.len();
+        let carry = if self_len < other_len {
+            let mid_carry = __add2(&mut self.data[..], &other.data[..self_len]);
+            self.data.extend_from_slice(&other.data[self_len..]);
+            __add2(&mut self.data[self_len..], &[mid_carry])
+        } else {
+            __add2(&mut self.data[..], &other.data[..])
+        };
         if carry != 0 {
             self.data.push(carry);
         }
```
This removed the loss, and then the changes in benchmark times due to the commit of the PR was simply within the margin of error, so it was inconclusive.

So now I'm not sure about the PR. What I think is this:
1. The patch in this comment can be applied anyway to very significantly improve the speed of short_val + long_ref. I could do this in another PR if that is desired.
2. If we want to slightly optimize ref + ref, which was the original intention of the PR, we can do it on a case-by-case basis. For example for `BigUint` the only case I think would get any benefit is `BitAnd`, and in that case a special implementation could be beneficial anyway because there even the capacity could be the shorter length (not long capacity with short length). If desired, I could do another PR to reduce the capacity of the result of `&BigUint & &BigUint`.
3. This PR can be closed.

tspiteri:
I've created two PRs #34 for point 1 above and #35 for point 2.

PR #34, opened 2018-03-03
do not zero-extend for addition when val is shorter than ref

tspiteri:
For `short_val + long_ref`, `long_ref + short_val` and `long_ref - short_val`, the current code extends the short value with zeros and then performs the operation. Instead, this commit splits the addition/subtraction into two parts, the lower and the higher parts. After performing the addition/subtraction on the low part, the short value is extended with the high part of the long reference, and the intermediate carry/borrow is then added to/subtracted from the high part.


cuviper:
Looks good, thanks!

bors r+

PR #35, opened 2018-03-03
avoid over-allocation for bitwise and of references

tspiteri:
For `&BigUint & &BigUint`, clone the smaller value to avoid over-allocation, as the result will have a length less than or equal to the minimum length of the operands.


cuviper:
Looks good, thanks!

bors r+

PR #39, opened 2018-05-02
Make it explicit that serialization is in u32

cuviper:
Add explicit type annotations to make sure we serialize `BigUint` like a
`Vec<u32>`.  If we ever change the internal representation, we still
need to remain compatible with serialized data.

Also, deserializing `BigUint` and `BigInt` now uses their normal
constructor methods, to make sure they are normalized.  We shouldn't
assume that all incoming data is already well-formed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,9 @@ version = "0.4"
 optional = true
 version = "1.0"
 
+[dev-dependencies]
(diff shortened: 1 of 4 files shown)


cuviper:
Now it's verifying with `serde_test` too.

bors r+

PR #42, opened 2018-05-14
Add a special case for single-digit divisors

cuviper:
It was pointed out in the blog post [Big Integers in Zig] that we don't
have a special case in `num-bigint` for single-digit divisors.  While
you can already get this optimization by dividing directly by `u32`,
it's easy to make small `BigUint` divisors work like this too.

    $ cargo benchcmp baseline single-div
     name                   baseline ns/iter  single-div ns/iter  diff ns/iter   diff %  speedup
     factorial_div_biguint  5,638,353         1,005,488             -4,632,865  -82.17%   x 5.61

`BigInt` will also gain from this, since it uses `BigUint` division
internally.

Running [zig-bn's facdiv-rs] shows a nice improvement too.  My i7-7700K
with Rust 1.26 goes from 4.15 seconds to just 0.65 -- a 6.38x speedup!

[Big Integers in Zig]: https://tiehuis.github.io/big-integers-in-zig#division-test-single-limb
[zig-bn's facdiv-rs]: https://github.com/tiehuis/zig-bn/tree/master/bench/facdiv/crate-facdiv-rs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,9 @@ readme = "README.md"
 name = "bigint"
 
 [[bench]]
+name = "factorial"
+
+[[bench]]
 name = "gcd"
 
 [[bench]]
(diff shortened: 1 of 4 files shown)


cuviper:
Note, this is on `num-bigint-0.1.x` rather than `master`, so I can publish it sooner for 0.1 users.  I'll do a followup to merge this into `master`.

PR #51, opened 2018-05-30
Implement sqrt for BigInt and BigUint

mancabizjak:
Related to [#3](https://github.com/rust-num/num-integer/issues/3).

I needed to calculate square root of a big integer in the project I'm working on, so I thought maybe this can be useful to others as well. 

This PR adds `sqrt()` method for calculating square root of a non-negative integer to `BigInt` and `BigUint` types. Square root is obtained via Newton's method, with a reasonable initial guess to improve performance.

Benchmark is included, although I had to replace the implementation of `StdRng` with `ThreadRng` in order to be able to run it.

### DIFF
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -1756,7 +1756,6 @@ impl BigUint {
     /// square root of `self`.
     pub fn sqrt(&self) -> Self {
         let one = BigUint::one();
-        let two = BigUint::from(2 as u8);
 
         // Trivial cases
         if self.is_zero() {
@@ -1775,18 +1774,18 @@ impl BigUint {
         // Set initial guess to something >= floor(sqrt(self)), but as low
         // as possible to speed up convergence.
         let bit_len = self.len() * big_digit::BITS;
-        let guess = one.shl(bit_len/2 + 1);
+        let guess = one << (bit_len/2 + 1);
 
         let mut u = guess;
         let mut s: BigUint;
 
         loop {
             s = u;
-            let q = self.div_floor(&s);
-            let t: BigUint = &s + &q;
+            let q = self / &s;
+            let t: BigUint = &s + q;
 
             // Compute the candidate value for next iteration
-            u = t.div_floor(&two);
+            u = t >> 1;
 
             if u >= s { break; }
         }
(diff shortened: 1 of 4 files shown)


cuviper:
> Benchmark is included, although I had to replace the implementation of StdRng with ThreadRng in order to be able to run it.

Ah, sorry, I didn't update the benchmarks for the new `rand`.  I'll do a separate PR to clean that up.

src/biguint.rs
```diff
@@ -1749,6 +1749,50 @@ impl BigUint {
...
+        let bit_len = self.len() * big_digit::BITS;
+        let guess = one.shl(bit_len/2 + 1);
+
+        let mut u = guess;
+        let mut s: BigUint;
+
+        loop {
+            s = u;
+            let q = self.div_floor(&s);
+            let t: BigUint = &s + &q;
```
cuviper: We don't need `q` after this, so adding `&s + q` by value will let it reuse that memory allocation.

src/biguint.rs
```diff
@@ -1749,6 +1749,50 @@ impl BigUint {
...
+
+            // Compute the candidate value for next iteration
+            u = t.div_floor(&two);
```
cuviper: Any reason you're using `div_floor` instead of the `/` operator?  With the operator you can also use integer literals, like `t / 2u32`.  But even more, `t >> 1` is probably faster.

src/biguint.rs
mancabizjak: Indeed, fixed.

src/biguint.rs
mancabizjak: No reason whatsoever, it just slipped my mind it seems. I replaced the usage of `div_floor` with the `/` and `>>` operators as appropriate, and also replaced the `lsh` call with `<<`. Looks much cleaner now.

mancabizjak:
Great, I'd be happy to. I already have a draft  for the `nth_root` implementation but it needs a little polishing. Hopefully I'll have everything ready some time the following week.

mancabizjak:
Closing in favor of #56.

PR #53, opened 2018-05-31
switch gen_biguint to fill_bytes

TheIronBorn:
Changes `gen_biguint` from a `push(gen::<u32>)` method to rand's [`fill_bytes`](https://docs.rs/rand/0.5.0/rand/trait.RngCore.html#tymethod.fill_bytes). This should improve performance in most cases.

- For small PRNGs which only natively generate 64 bits (like Xorshift64 or [`splitmix64.c`](http://prng.di.unimi.it/splitmix64.c)), this will no longer throw away half the bits generated. 
- For block PRNGs like `StdRng`, this should reduce overhead.
- For an SIMD PRNG (https://github.com/rust-lang-nursery/rand/issues/377), this would be a significant improvement.

```diff,ignore
 name         no_fill ns/iter  fill ns/iter  diff ns/iter   diff %  speedup 
+rand_1009    256              222                    -34  -13.28%   x 1.15 
+rand_131072  27,366           14,715             -12,651  -46.23%   x 1.86 
+rand_2048    459              357                   -102  -22.22%   x 1.29 
-rand_256     93               130                     37   39.78%   x 0.72 
+rand_4096    842              557                   -285  -33.85%   x 1.51 
-rand_64      69               92                      23   33.33%   x 0.75 
+rand_65536   13,625           7,382               -6,243  -45.82%   x 1.85 
+rand_8192    1,836            869                   -967  -52.67%   x 2.11
```
(i.e. `rand_1009` does `gen_biguint(1009)`. All benches are powers of two except `rand_1009`)

(Let me know if you want the `rand_` benches added)

### DIFF
--- a/src/bigrand.rs
+++ b/src/bigrand.rs
@@ -2,6 +2,7 @@
 
 use rand::prelude::*;
 use rand::distributions::uniform::{SampleUniform, UniformSampler};
+use rand::AsByteSliceMut;
 
 use BigInt;
 use BigUint;
@@ -39,13 +40,15 @@ impl<R: Rng + ?Sized> RandBigInt for R {
     fn gen_biguint(&mut self, bit_size: usize) -> BigUint {
         use super::big_digit::BITS;
         let (digits, rem) = bit_size.div_rem(&BITS);
-        let mut data = Vec::with_capacity(digits + 1);
-        for _ in 0..digits {
-            data.push(self.gen());
-        }
+        let mut data = vec![BigDigit::default(); digits + (rem > 0) as usize];
+        // `fill_bytes` is faster than many `gen::<u32>` calls
+        self.fill_bytes(data[..].as_byte_slice_mut());
+        // Swap bytes per the `Rng::fill` source. This might be
+        // unnecessary if reproducibility across architectures is not
+        // desired.
+        data.to_le();
         if rem > 0 {
-            let final_digit: BigDigit = self.gen();
-            data.push(final_digit >> (BITS - rem));
+            data[digits] >>= BITS - rem;
         }
         BigUint::new(data)
     }
--- a/benches/bigint.rs
+++ b/benches/bigint.rs
@@ -214,6 +214,52 @@ fn from_str_radix_36(b: &mut Bencher) {
     from_str_radix_bench(b, 36);
 }
 
+fn rand_bench(b: &mut Bencher, bits: usize) {
+    let mut rng = get_rng();
+
+    b.iter(|| rng.gen_bigint(bits));
+}
+
+#[bench]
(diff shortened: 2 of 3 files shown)


src/bigrand.rs
```diff
@@ -39,13 +40,15 @@ impl<R: Rng + ?Sized> RandBigInt for R {
...
-        for _ in 0..digits {
-            data.push(self.gen());
-        }
+        let mut data = vec![BigDigit::default(); digits + (rem > 0) as usize];
+        // `fill_bytes` is faster than many `gen::<u32>` calls
+        self.fill_bytes(data[..].as_byte_slice_mut());
+        // Swap bytes per the `Rng::fill` source. This might be
+        // unnecessary if reproducibility across architectures is not
+        // desired.
+        data.to_le();
```
cuviper: This is an interesting question -- what guarantees should we make about reproducibility?  Is this PR a breaking change altogether, since seeded RNGs may now produce a different `BigUint`?  Maybe that's going too far to consider this across crate versions, but if we want it across architectures, we'll also need to be careful in the future about conditional `BigDigit` sizes, for instance.  (Right now this is using the public `BigUint::new` though, which should always be `u32`-based even if the internals change.)

TheIronBorn:
I don't know why the `bors` check is hanging.

cuviper:
No, it's just waiting for me to approve it.  Did you have any more thoughts about what (if anything) we should guarantee, when producing numbers from a seeded rng?

Also, yes, adding the benches would be nice.

TheIronBorn:
I don't really use the crate, I just happened to see this implementation could be improved. So I'm not sure.

src/bigrand.rs
dhardy: This may be a value-breaking change (depending on the PRNG used), but I think will usually not be. (Rand does not specify that `fill_bytes` must output the same bits in the same order as repeated calls to `next_u32`, though I think this is the case for all current generators.)

TheIronBorn:
You could add a deprecated function using the old method. I’m not sure how useful that would be without implementing all the surrounding code though.

cuviper:
OK, I've added some tests of value stability.  I generated the expected values on the current master branch, and they do still match on this branch.  I think we should go ahead and make this change, and mention the caveat about *possible* effects on reproducibility (depending on PRNG) in the release notes.

Before I merge, I'm going to see if I can run this on Fedora's infrastructure with all architectures, but I need to update a few dependencies there first -- including `rand` itself!

cuviper:
All arches passed, including big-endian ppc64 and s390x.
https://koji.fedoraproject.org/koji/taskinfo?taskID=28044004

Thanks!

bors r+

Issue #29, opened 2018-02-27
Implement Pow for BigUint and BigInt


clarcharr:
See rust-num/num-traits#38 for details on this trait.

vks:
Is this necessary given that it is implemented in `num-traits` now?

cuviper:
What do you mean?  `Pow` is defined in `num-traits`, and implemented for the primitive types, but we need to implement it here for bigints.

vks:
I thought `num-traits` provided a generic implementation, but apparently that is not the case.

thomwiggers:
Let's see if I can do this – I need in anyhow.

vks:
@cuviper I found it: [num_traits::pow](https://docs.rs/num-traits/0.2.5/num_traits/pow/fn.pow.html). This should work for `BigInt` as well. (Something like `powm` in [this `num-integer` PR](https://github.com/rust-num/num-integer/pull/10) would probably be more efficient though.)

cuviper:
> I found it: num_traits::pow. This should work for BigInt as well.

Yes, but that only works by-value with a `usize` exponent.  The `Pow` trait is more flexible.

> (Something like powm in this num-integer PR would probably be more efficient though.)

Modular `pow` is a different thing though -- and FWIW we do have inherent `modpow` methods here.

vks:
> Yes, but that only works by-value with a usize exponent. The Pow trait is more flexible.

A `u32` exponent should be enough, and by-value is more efficient.

windelbouwman:
I'm interested in this pow function, will there be a new release of this crate at some point containing pow, or is there work left to do?

cuviper:
Thanks for the reminder - I've published num-bigint 0.2.1.

peterbraden:
Still not updated in num? I get:
`    Updating crates.io index
 Downloading num v0.2.0
 Downloading num-bigint v0.2.1
 Downloading num-rational v0.2.1
   Compiling num-integer v0.1.39
   Compiling num-iter v0.1.37
   Compiling num-complex v0.2.1
   Compiling num-bigint v0.2.1
   Compiling num-rational v0.2.1
   Compiling num-bigint v0.1.40
error[E0277]: the trait bound `biguint::BigUint: num_traits::Num` is not satisfied`

cuviper:
[...]

PR #54, opened 2018-07-04
Implement Pow for BigUint and BigInt

thomwiggers:
Implements #29

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -838,6 +838,15 @@ macro_rules! pow_impl {
                 BigInt::from_biguint(powsign(self.sign, &rhs), (&self.data).pow(rhs))
             }
         }
+
+        impl<'a, 'b> Pow<&'b $type> for &'a BigInt {
+            type Output = BigInt;
+
+            #[inline]
+            fn pow(self, rhs: &$type) -> BigInt {
+                BigInt::from_biguint(powsign(self.sign, rhs), (&self.data).pow(rhs))
+            }
+        }
     }
 }
 
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -66,24 +66,6 @@ impl Mul<Sign> for Sign {
     }
 }
 
-impl<'a, T: Integer> Pow<&'a T> for Sign {
-    type Output = Sign;
-
-    #[inline]
-    fn pow(self, other: &T) -> Sign {
-        if other.is_zero() {
-            Plus
-        } else if self != Minus {
-            self
-        } else if other.is_odd() {
-            self
-        } else {
-            -self
-        }
-    }
-}
-
-
 #[cfg(feature = "serde")]
 impl serde::Serialize for Sign {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
@@ -830,6 +812,22 @@ impl Signed for BigInt {
 }
 
 
+/// Help function for pow
+///
+/// Computes the effect of the exponent on the sign.
+#[inline]
+fn powsign<T: Integer>(sign: Sign, other: &T) -> Sign {
+    if other.is_zero() {
+        Plus
+    } else if sign != Minus {
+        sign
+    } else if other.is_odd() {
+        sign
+    } else {
+        -sign
+    }
+}
+
 macro_rules! pow_impl {
     ($type:ty) => {
         impl<'a> Pow<$type> for &'a BigInt {
@@ -837,7 +835,7 @@ macro_rules! pow_impl {
 
             #[inline]
             fn pow(self, rhs: $type) -> BigInt {
-                BigInt::from_biguint(self.sign.pow(&rhs), self.data.pow(rhs))
+                BigInt::from_biguint(powsign(self.sign, &rhs), (&self.data).pow(rhs))
             }
         }
     }
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -18,7 +18,7 @@ use serde;
 
 use integer::{Integer, Roots};
 use traits::{ToPrimitive, FromPrimitive, Num, CheckedAdd, CheckedSub,
-             CheckedMul, CheckedDiv, PrimInt, Signed, Zero, One, Pow};
+             CheckedMul, CheckedDiv, Signed, Zero, One, Pow};
 
 use self::Sign::{Minus, NoSign, Plus};
 
@@ -66,16 +66,16 @@ impl Mul<Sign> for Sign {
     }
 }
 
-impl<T: PrimInt + BitAnd<Output=T>> Pow<T> for Sign {
+impl<'a, T: Integer> Pow<&'a T> for Sign {
     type Output = Sign;
 
     #[inline]
-    fn pow(self, other: T) -> Sign {
-        if self != Minus {
-            self
-        } else if other == T::zero() {
+    fn pow(self, other: &T) -> Sign {
+        if other.is_zero() {
             Plus
-        } else if other & T::one() == T::one() {
+        } else if self != Minus {
+            self
+        } else if other.is_odd() {
             self
         } else {
             -self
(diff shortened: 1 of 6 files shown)


src/bigint.rs
```diff
@@ -66,6 +66,24 @@ impl Mul<Sign> for Sign {
...
 
+impl<T: PrimInt + BitAnd<Output=T>> Pow<T> for Sign {
+    type Output = Sign;
+
+    #[inline]
+    fn pow(self, other: T) -> Sign {
+        if self != Minus {
+            self
+        } else if other == T::zero() {
+            Plus
```
cuviper: The check should use `other.is_zero()`, and this case should be first, because we also need `Plus` for `NoSign.pow(0)`.

src/bigint.rs
```diff
@@ -66,6 +66,24 @@ impl Mul<Sign> for Sign {
...
+        } else if other & T::one() == T::one() {
```
cuviper: Slightly better to use (other & T::one()).is_one() -- or if you switch to `T: Integer`, it's just `other.is_odd()`.

src/bigint.rs
```diff
@@ -811,6 +829,28 @@ impl Signed for BigInt {
     }
 }
 
+impl<T: PrimInt> Pow<T> for BigInt
+    where BigUint: Pow<T, Output=BigUint>
```
cuviper: Let's have exponent references too, as I mentioned on `BigUint`.

I think it's *technically* a breaking change to do this generically, as someone could already have an impl, however silly, like:

```rust
struct Foo(usize);

impl Pow<Foo> for BigInt {
    type Output = BigInt;
    
    fn pow(self, exp: Foo) -> BigInt {
        num_traits::pow(self, exp.0)
    }
}
```

Let's just avoid that question by implementing it in macros like `BigUint`.

src/biguint.rs
```diff
@@ -434,6 +434,57 @@ impl One for BigUint {
...
 
+macro_rules! pow_impl {
+    ($unsigned:ty, $signed:ty) => {
+        impl pow::Pow<$unsigned> for BigUint {
+            type Output = BigUint;
+
+            #[inline]
+            fn pow(self, exp: $unsigned) -> Self::Output {
+                // square and multiply from num_traits::pow
+                pow::pow(self, exp as usize)
```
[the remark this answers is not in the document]
thomwiggers: The exponents we implement are all `Copy` so I don't see much value for a reference – one which we'll immediately clone anyway, as we need to mutate it.

src/biguint.rs
```diff
@@ -434,6 +434,57 @@ impl One for BigUint {
...
+            }
+        }
+
+        impl<'a> pow::Pow<$unsigned> for &'a BigUint {
+            type Output = BigUint;
+
+            #[inline]
+            fn pow(self, exp: $unsigned) -> Self::Output {
+                // square and multiply from num_traits::pow
+                pow::pow(self.clone(), exp as usize)
```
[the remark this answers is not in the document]
thomwiggers: This also immediately deals with the `u128` implementation that was still `FIXME`.

src/biguint.rs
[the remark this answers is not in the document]
thomwiggers: I also got rid of the implementation on `BigUint` and `BigInt` directly, the references should suffice.

tests/bigint.rs
```diff
@@ -1092,3 +1092,29 @@ fn test_iter_product_generic() {
...
     assert_eq!(result, data.into_iter().product());
 }
+
+#[test]
+fn test_pow() {
+    let one = 1u32.to_bigint().unwrap();
+    let two = 2u32.to_bigint().unwrap();
+    let minus_two = -2i32.to_bigint().unwrap();
+    let four = 4u32.to_bigint().unwrap();
+    let eight = 8u32.to_bigint().unwrap();
```
cuviper: These can use `From` like `BigInt::from(8u32)` nowadays -- although I know many other tests are still doing it the hard way too. (We ought to clean that up...)

src/biguint.rs
```diff
@@ -434,6 +434,47 @@ impl One for BigUint {
...
+    ($unsigned:ty) => {
+         impl<'a> pow::Pow<$unsigned> for &'a BigUint {
+            type Output = BigUint;
+
+            #[inline]
+            fn pow(self, mut exp: $unsigned) -> Self::Output {
+                // square and multiply from num_traits::pow in the next branch
+                let mut base = self.clone();
+
+                if exp == 0 { return BigUint::one(); }
```
cuviper: Might as well check this first, to potentially avoid a useless clone.

src/biguint.rs
```diff
@@ -434,6 +434,47 @@ impl One for BigUint {
...
 impl Unsigned for BigUint {}
 
+macro_rules! pow_impl {
+    ($unsigned:ty) => {
+         impl<'a> pow::Pow<$unsigned> for &'a BigUint {
+            type Output = BigUint;
+
+            #[inline]
+            fn pow(self, mut exp: $unsigned) -> Self::Output {
+                // square and multiply from num_traits::pow in the next branch
```
cuviper: Let's not immortalize the next branch in comments -- that stuff is obsolete.  If you're trying to give credit, then a note in your commit message is enough.

src/bigint.rs
```diff
@@ -66,6 +66,24 @@ impl Mul<Sign> for Sign {
     }
 }
 
+impl<'a, T: Integer> Pow<&'a T> for Sign {
```
cuviper: I'm still hesitating whether we really want a public `impl` for this.  It's kind of silly to have a bunch of traits implemented for `Sign`.  (An alternative is to just have a private `fn` to get the new sign.)

src/biguint.rs
```diff
@@ -434,6 +434,47 @@ impl One for BigUint {
...
+                if exp == 0 { return BigUint::one(); }
+
+                while exp & 1 == 0 {
+                    base = &base * &base;
+                    exp >>= 1;
+                }
+
+                if exp == 1 { return base; }
+
+                let mut acc = self.clone();
```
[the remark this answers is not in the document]
cuviper: Err, actually you do have a test for `pow(3)` -- but since that has no *trailing* zeros, `base` and `self` are still the same.  Try an exponent like 10 to see the failure I'm talking about.

src/bigint.rs
thomwiggers: Private fn :heavy_check_mark:

src/biguint.rs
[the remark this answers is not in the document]
thomwiggers: oops, I missed that when copying the pow implementation from `next`.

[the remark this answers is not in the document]
thomwiggers: Added a few tests as well.

tests/bigint.rs
thomwiggers: :heavy_check_mark:

src/bigint.rs
thomwiggers: I'd have expected the entire `Sign` trait private by the way, but there's probably some good reason for it.

src/biguint.rs
thomwiggers: Good point.

src/biguint.rs
thomwiggers: :heavy_check_mark:

src/biguint.rs
```diff
@@ -434,6 +434,55 @@ impl One for BigUint {
 
 impl Unsigned for BigUint {}
 
+macro_rules! pow_impl {
+    ($T:ty) => {
+        impl<'a> Pow<$T> for &'a BigUint {
```
cuviper: Just for posterity: I was going to ask you to add `for BigUint` (by value) too, but I tried it myself first.  That makes it really annoying when you do have a bigint value, as the `pow` method will then try to move ownership.  You'd have to manually reference `(&i).pow(exp)` to avoid this and get the by-ref method, which is really bad for ergonomics.

So if anyone asks why we don't have that `impl`, here's the reason.

src/bigint.rs
cuviper: It's an `enum Sign`, not a trait, and it's a public argument for some of the constructors.  We have to indicate the sign *somehow*, since we're not using 2's complement, so an `enum` is relatively clean.  It usually stays out of the way. :)

thomwiggers:
Yeah, that's something I've run into as well. I probably meant to comment that but forgot :p

With some luck will the compiler take care of the overhead of the ref where it could've been by value.

Josh Stone <notifications@github.com> schreef op August 4, 2018 12:12:04 AM UTC:
>cuviper commented on this pull request.
>
>
>
>> @@ -434,6 +434,55 @@ impl One for BigUint {
> 
> impl Unsigned for BigUint {}
> 
>+macro_rules! pow_impl {
>+    ($T:ty) => {
>+        impl<'a> Pow<$T> for &'a BigUint {
>
>Just for posterity: I was going to ask you to add `for BigUint` (by
>value) too, but I tried it myself first.  That makes it really annoying
>when you do have a bigint value, as the `pow` method will then try to
>move ownership.  You'd have to manually reference `(&i).pow(exp)` to
>avoid this and get the by-ref method, which is really bad for
>ergonomics.
>
>So if anyone asks why we don't have that `impl`, here's the reason.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/rust-num/num-bigint/pull/54#pullrequestreview-143364630

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #56, opened 2018-07-11
Implement Roots for BigInt and BigUint

mancabizjak:
Supersedes #51 .

Since there is now a `Roots` trait with `sqrt`, `cbrt` and `nth_root` methods in the `num-integer` crate, this PR implements it for `BigInt` and `BigUint` types. I also added inherent methods on both types to allow the users access to all these functions without having to import `Roots`.

PS: `nth_root` currently  uses `num_traits::pow`. Should we perhaps wait for #54 to get merged, and then replace the call to use the new `pow::Pow` implementation on `BigUint`?

### DIFF
--- a/benches/bigint.rs
+++ b/benches/bigint.rs
@@ -4,13 +4,14 @@
 extern crate test;
 extern crate num_bigint;
 extern crate num_traits;
+extern crate num_integer;
 extern crate rand;
 
 use std::mem::replace;
 use test::Bencher;
 use num_bigint::{BigInt, BigUint, RandBigInt};
 use num_traits::{Zero, One, FromPrimitive, Num};
-use rand::{SeedableRng, StdRng};
+use rand::{SeedableRng, StdRng, Rng};
 
 fn get_rng() -> StdRng {
     let mut seed = [0; 32];
@@ -341,4 +342,33 @@ fn modpow_even(b: &mut Bencher) {
     let m = BigUint::from_str_radix(RFC3526_2048BIT_MODP_GROUP, 16).unwrap() - 1u32;
 
     b.iter(|| base.modpow(&e, &m));
+}
+
+#[bench]
+fn roots_sqrt(b: &mut Bencher) {
+    let mut rng = get_rng();
+    let x = rng.gen_biguint(2048);
+
+    b.iter(|| x.sqrt());
+}
+
+#[bench]
+fn roots_cbrt(b: &mut Bencher) {
+    let mut rng = get_rng();
+    let x = rng.gen_biguint(2048);
+
+    b.iter(|| x.cbrt());
+}
+
+#[bench]
+fn roots_nth(b: &mut Bencher) {
+    let mut rng = get_rng();
+    let x = rng.gen_biguint(2048);
+    // Although n is u32, here we limit it to the set of u8 values since it
+    // hugely impacts the performance of nth_root due to exponentiation to
+    // the power of n-1. Using very large values for n is also not very realistic,
+    // and any n > x's bit size produces 1 as a result anyway.
+    let n: u8 = rng.gen();
+
+    b.iter(|| { x.nth_root(n as u32) });
 }
--- a/benches/bigint.rs
+++ b/benches/bigint.rs
@@ -11,7 +11,7 @@ use std::mem::replace;
 use test::Bencher;
 use num_bigint::{BigInt, BigUint, RandBigInt};
 use num_traits::{Zero, One, FromPrimitive, Num};
-use rand::{SeedableRng, StdRng, Rng};
+use rand::{SeedableRng, StdRng};
 
 fn get_rng() -> StdRng {
     let mut seed = [0; 32];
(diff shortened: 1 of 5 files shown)


src/biguint.rs
```diff
@@ -1026,6 +1026,52 @@ impl Integer for BigUint {
...
+        if self.is_zero() {
+            return BigUint::zero();
+        }
+
+        if self.is_one() {
+            return one;
+        }
+
+        let n = n as usize;
+        let n_min_1 = (n as usize) - 1;
```
cuviper: Don't need to cast `n as usize` twice.

src/biguint.rs
```diff
@@ -1749,6 +1795,25 @@ impl BigUint {
         }
         acc
     }
+
+    /// Returns the truncated principal square root of `self` --
+    /// see [Roots::sqrt](Roots::sqrt).
+    // struct.BigInt.html#trait.Roots
```
cuviper: These links don't look right -- did you forget to complete this?

src/biguint.rs
```diff
@@ -1026,6 +1026,52 @@ impl Integer for BigUint {
     }
 }
 
+impl Roots for BigUint {
+    fn nth_root(&self, n: u32) -> Self {
```
cuviper: I think it would probably be valuable to have dedicated `sqrt` and `cbrt` implementations -- did you try it?  Maybe benchmarks would prove me wrong, but we can also wait until later.

tests/roots.rs
```diff
@@ -0,0 +1,84 @@
...
+extern crate num_integer;
+extern crate num_traits;
+
+mod biguint {
+    use num_bigint::BigUint;
+    use num_traits::FromPrimitive;
+    use std::str::FromStr;
+
+    fn check(x: i32, n: u32, expected: i32) {
+        let big_x: BigUint = FromPrimitive::from_i32(x).unwrap();
```
cuviper: If you use an unsigned type instead of `i32`, then `From` makes this simpler -- `BigUint::from(x)`.

tests/roots.rs
```diff
@@ -0,0 +1,84 @@
...
+        assert_eq!(x.nth_root(10), expected);
+    }
+}
+
+mod bigint {
+    use num_bigint::BigInt;
+    use num_traits::FromPrimitive;
+
+    fn check(x: i32, n: u32, expected: i32) {
+        let big_x: BigInt = FromPrimitive::from_i32(x).unwrap();
```
cuviper: Similarly, `BigInt::from(x)` should be fine.

benches/bigint.rs
```diff
@@ -342,3 +343,32 @@ fn modpow_even(b: &mut Bencher) {
...
+
+#[bench]
+fn roots_nth(b: &mut Bencher) {
+    let mut rng = get_rng();
+    let x = rng.gen_biguint(2048);
+    // Although n is u32, here we limit it to the set of u8 values since it
+    // hugely impacts the performance of nth_root due to exponentiation to
+    // the power of n-1. Using very large values for n is also not very realistic,
+    // and any n > x's bit size produces 1 as a result anyway.
+    let n: u8 = rng.gen();
```
cuviper: This is from a seeded RNG, so what value do you get?  I think you might as well just pick a reasonable `n` directly.  Random `x` is still fine.

src/biguint.rs
mancabizjak: Here are the benchmark results:
```bash
test roots_cbrt            ... bench:      48,670 ns/iter (+/- 1,322)
test roots_cbrt_nth_root_3 ... bench:      52,285 ns/iter (+/- 1,541)
test roots_sqrt            ... bench:      53,752 ns/iter (+/- 4,954)
test roots_sqrt_nth_root_2 ... bench:      57,622 ns/iter (+/- 1,585)
```
`roots_cbrt` and `roots_sqrt` test the optimized versions - I took `sqrt` from #51 and adjusted it for `cbrt`. The other two benchmarks just called `nth_root(2)` and `nth_root(3)` - of course, for the sake of benchmark, I didn't forward the calls to `nth_root` with `n=2` and `n=3` to the optimized `sqrt` and `cbrt`, but I'll definitely do that, since I see this is how it's done in impls for primitive types as well.

I would prefer to incorporate the dedicated implementations in this PR, although I'm struggling a little since I'm not particularly fond of having  parts of `nth_root`, `sqrt`, `cbrt` replicated in all the methods. But I'm not too sure what (if anything) would be best to do about it :slightly_smiling_face:

src/biguint.rs
mancabizjak: Eh, forgot to remove.

mancabizjak:
Thanks for the review! I overrided `sqrt` and `cbrt` methods with versions slightly more performant than  their default implementations with `nth_root`. Recurring checks for trivial cases are now a bit more compact.

In addition, after glancing at the test code for implementations of `Roots` on `PrimInt`s, I felt that the test code was not thorough enough, so I improved it a little.

I also fixed some inconsistencies in comments and made panic messages a bit clearer.

src/bigint.rs
```diff
@@ -2538,6 +2557,25 @@ impl BigInt {
         };
         BigInt::from_biguint(sign, mag)
     }
+
+    /// Returns the truncated principal square root of `self` --
+    /// see [Roots::sqrt](Roots::sqrt).
+    // struct.BigInt.html#trait.Roots
```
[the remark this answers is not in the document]
mancabizjak: When I run `cargo doc --no-deps --open` and navigate to the page of `BigInt` or `BigUint` structs, paths of the form `Roots::sqrt` are rendered as links to `https://docs.rs/num-integer/0.1/num_integer/trait.Roots.html#method.sqrt` - I'm a little confused, is this not what we want? I was merely following [this RFC](https://github.com/rust-lang/rfcs/blob/master/text/1946-intra-rustdoc-links.md), but I could be missing something.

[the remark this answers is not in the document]
cuviper: Ah, it seems that works on nightly, but not beta or stable rustc yet.  We should make those explicit for now.

cuviper:
Thanks!

bors r+

PR #62, opened 2018-07-21
Finish i128 support

### LINKED ISSUE #40: Add support for 128-bit primitives
Rust 1.26 is stabilizing `i128` and `u128`.  We should add support for for these, to the same extent that other primitive types are supported.  This should be a conditional feature, rather than forcing all of `num-bigint` up to the very latest compiler -- e.g. rust-num/num-traits#60 has a new feature "i128".

- [x] convert to and from 128-bit integers #44 
- [x] implement operators with 128-bit operands
- [ ] implement `BigDigit = u64` with `DoubleBigDigit = u128`
  - still conditional on new enough rustc, and probably only for specific target CPUs

dignifiedquire:
Depends on #63 and #64 

Implements `BigDigit = u64` ~~behind a feature flag `u64_digit`~~ for all 64-bit targets with `u128`.

TODOs
- [x] find a way to make the new methods that take `BigDigit`s for constructing `BigUint`s private
- [x] Fix `modpow`, the last broken test.

Closes #40


cuviper:
Thanks for taking this on! It will be a few days before I can review this in depth, but a few notes:

- Running rustfmt is a good thing to do, but I definitely want that in a separate PR where it's clear that no code is actually changing. I recently ran `cargo fmt` on most of the other num crates, but I held off here because I didn't want to conflict with outstanding PRs.

- Everything should use `cfg(has_i128)` rather than the feature, so auto-detection just works.

- We need benchmark results, across multiple platforms -- at least 32-bit and 64-bit. I have a feeling we'll want to choose larger `BigDigit` only for some platforms. And even on 64-bit, I'm worried in particular that 128-bit division will be slow.

- If we do use different sizes for different targets, it's probably best to decide that in the build script, and output another simple flag akin to `has_i128`.

- I explicitly don't want methods with `BigDigit` in the public API. This was specifically cleaned up in 0.2 to prepare for supporting different sizes. IMO it's too much of a portability footgun for someone to write code that works fine on their main target, but becomes totally wrong on a different target.

- I think it makes more sense to add the basic op support as it's own PR, because that doesn't have as many questions or design decisions.

dignifiedquire:
Thanks for the quick answer, I'll pull things into their own PRs, first one is here #63

dignifiedquire:
> I explicitly don't want methods with BigDigit in the public API. This was specifically cleaned up in 0.2 to prepare for supporting different sizes. IMO it's too much of a portability footgun for someone to write code that works fine on their main target, but becomes totally wrong on a different target.

I understand, but `BigInt` needs to be able to construct a `BigUint` from `BigDigit`s, and they are not in the same module currently. So what would your preferred approach be for that?

dignifiedquire:
> Everything should use cfg(has_i128) rather than the feature, so auto-detection just works.

Done for all the operations in #64

dignifiedquire:
> If we do use different sizes for different targets, it's probably best to decide that in the build script, and output another simple flag akin to has_i128.

I have introduced a new feature `u64_digit`, to guard the switch for now. I was thinking it could be shipped with the default being disabled, and then gather benchmarks and feedback on how it is working for a while, before turning it on by default on some platforms.

cuviper:
This needs a rebase after #63 and #64 landed.  Were you ever able to improve the division performance, as we discussed in #40?  If you could at least update the rest of what you've done, perhaps I can look at that too.

dignifiedquire:
- I'll do a rebase later today
- I got modpow working, by switching to a different algorithm.
- I got the perf only under control by using inline assembly, which you probably don't want to merge, so that will need another solution

cuviper:
OK, no rush, I'm just trying to get around to things I've been neglecting. :)

dignifiedquire:
@cuviper I rebased this, and it is passing tests, (except serialization with `u64` digits, I missed that earlier, just marked as `unimplemented` for now)

When you have some time, it would be great if we can work through the remaining issues to figure out how to get this ready for merge.

cuviper:
@dignifiedquire -- If you're still around, I think I've reconciled the performance losses well enough, and I've made quite a few other cleanups as well. A counter-review would be appreciated.

cc @maxbla -- since you've been hacking on num lately, maybe you'd like to review this too?

I have no doubt there are more cleanup opportunities, but I hope it's Good Enough that we can merge for the performance benefits at hand.

maxbla:
I'd be happy take a look. A comprehensive review might take some time - this PR seems fairly substantial.

cuviper:
I'll appreciate whatever level of detail you can manage, thanks!

cuviper:
Rebased to resolve conflicts with the quickcheck PR.

src/algorithms.rs
```diff
@@ -68,26 +69,57 @@ fn div_wide(hi: BigDigit, lo: BigDigit, divisor: BigDigit) -> (BigDigit, BigDigi
...
+#[inline]
 pub fn rem_digit(a: &BigUint, b: BigDigit) -> BigDigit {
-    let mut rem: DoubleBigDigit = 0;
-    for &digit in a.data.iter().rev() {
-        rem = (rem << big_digit::BITS) + DoubleBigDigit::from(digit);
-        rem %= DoubleBigDigit::from(b);
+    let mut rem = 0;
+
+    if b <= big_digit::HALF {
+        for &digit in a.data.iter().rev() {
```
maxbla: I think this could use a `fold` instead, although I don't know how that would affect performance (I assume it would make no difference). Something like `a.data.iter().rev().fold(0, |rem, digit| div_half(rem, digit, b).1)`. This change could reduce this function's length to 4 lines.

src/algorithms.rs
```diff
@@ -68,26 +69,57 @@ fn div_wide(hi: BigDigit, lo: BigDigit, divisor: BigDigit) -> (BigDigit, BigDigi
...
+            let (_, r) = div_half(rem, digit, b);
+            rem = r;
+        }
+    } else {
+        for &digit in a.data.iter().rev() {
```
maxbla: same here

src/algorithms.rs
```diff
@@ -655,9 +692,9 @@ fn div_rem_core(mut a: BigUint, b: &BigUint) -> (BigUint, BigUint) {
         let mut prod = b * &q0;
 
         while cmp_slice(&prod.data[..], &a.data[j..]) == Greater {
-            let one: BigUint = One::one();
-            q0 = q0 - one;
-            prod = prod - b;
+            let one: BigDigit = 1;
```
maxbla: Does this library draw a distinction between `rust`'s `1` literal and the ring identity element? I think I remember @cuviper calling the later "logical one" at some point. Do we think it is worth implementing `num_traits::identities::One` on `BigDigit` just to make the point that these two ones are different? It might allow more freedom in the future, if we were to change the underlying type of `BigDigit`.

src/algorithms.rs
```diff
@@ -655,9 +692,9 @@ fn div_rem_core(mut a: BigUint, b: &BigUint) -> (BigUint, BigUint) {
...
+            q0 -= one;
```
maxbla: nit: combine this line with the previous

src/bigint.rs
```diff
@@ -111,12 +111,28 @@ impl<'de> serde::Deserialize<'de> for Sign {
...
 
 /// A big signed integer type.
-#[derive(Clone, Debug, Hash)]
+#[derive(Debug, Hash)]
 pub struct BigInt {
     sign: Sign,
     data: BigUint,
 }
 
+impl Clone for BigInt {
```
maxbla: What does this offer over deriving `Clone`? Sign is not `Clone`'d in this implementation, but it is `Copy`'d, and I believe those will compile down to the same thing, so I don't see how this affects runtime.

src/algorithms.rs
maxbla: Actually, I'm not entirely sure what BigDigit is in all cases. I think it is an alias for u32 or u64 depending on platform, but both of those already implement `One`.

src/bigint.rs
```diff
@@ -1850,8 +1872,14 @@ impl<'a, 'b> Rem<&'b BigInt> for &'a BigInt {
 
     #[inline]
     fn rem(self, other: &BigInt) -> BigInt {
-        let (_, r) = self.div_rem(other);
-        r
+        if let Some(other) = other.to_u32() {
+            self % other
+        } else if let Some(other) = other.to_i32() {
```
maxbla: Since `other` is unsigned (because it is a `BigInt`), how could `other.to_u32()` return `None` but `other.to_i32()` return `Some(_)`?

maxbla: i.e. I think you can delete lines 1877 and 1878

src/monty.rs
```diff
@@ -1,129 +1,221 @@
...
+                z = montgomery(&zz, &zz, m, mr.n0inv, num_words);
+            }
+            zz = montgomery(
+                &z,
+                &powers[(yi >> (big_digit::BITS - n)) as usize],
+                m,
+                mr.n0inv,
+                num_words,
+            );
+            ::std::mem::swap(&mut z, &mut zz);
```
maxbla: what's up with the preceding namespace qualifier `::`? I didn't know you could do this, and it seems unnecessary. Wouldn't `std::mem::swap` work?

src/monty.rs
```diff
@@ -1,129 +1,221 @@
...
+    let mut z = powers[0].clone();
+    z.data.resize(num_words, 0);
+    let mut zz = BigUint::zero();
+    zz.data.resize(num_words, 0);
+
+    // same windowed exponent, but with Montgomery multiplications
+    for i in (0..y.data.len()).rev() {
+        let mut yi = y.data[i];
+        let mut j = 0;
+        while j < big_digit::BITS {
```
maxbla: nit: This could be a for loop, like `for j in (0..big_digit::BITS).step_by(n)`. In past experience, step_by is suboptimal (for large n), but this appears to no longer be the case.

src/algorithms.rs
cuviper: `BigDigit` is an internal-only detail at this point, just `u32` or `u64`. Since it's a `type` alias, it behaves exactly as the primitive types, including that it accepts literals like `1`. So at this time, I think going through `One` is needless abstraction.

> It might allow more freedom in the future, if we were to change the underlying type of `BigDigit`.

I don't expect we'd ever change to something that's not a primitive integer, but if we do, we'll get a nice compiler error during the refactoring. :) (i.e. this wouldn't fail silently)

src/algorithms.rs
cuviper: The small advantage here is in the type ascription, but actually `1u32` should work just as well for arithmetic, even when `BigDigit` is `u64`. I'll try it.

src/bigint.rs
cuviper: The derived `Clone` doesn't specialize `clone_from`. `Vec` can reuse its allocation through that, but the default `Clone::clone_from` just clobbers the old value with a whole new `clone`. I'll add a comment.

src/bigint.rs
cuviper: Are you thinking of `BigUint`? `BigInt` can be negative...

src/monty.rs
cuviper: You can only do that since [Rust 1.30](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1300-2018-10-25):

> Using a external crate no longer requires being prefixed with `::`.

But anyway I prefer to `use std::mem` at the top, then call `mem::swap`. Will change.

src/monty.rs
cuviper: Well, `step_by` was only stabilized in Rust 1.28, so for now it's out of reach.

build.rs
```diff
@@ -1,14 +1,69 @@
...
         panic!("i128 support was not detected!");
     }
 
     autocfg::rerun_path(file!());
+
+    write_radix_bases().unwrap();
+}
+
+#[allow(unknown_lints, bare_trait_objects)]
+fn write_radix_bases() -> Result<(), Box<Error>> {
```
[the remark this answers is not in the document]
cuviper: Done.

src/algorithms.rs
cuviper: It shouldn't make a difference for performance, but `rustfmt` won't leave that on a single line, so your golfing doesn't actually work out. :wink:

cuviper:
@maxbla Thanks for your review!

I noticed that you had comments for almost every file except `biguint.rs`, which has a lot of changes -- did you miss this? GitHub "helpfully" hides it:
> Large diffs are not rendered by default.

src/bigint.rs
maxbla: Yeah, you're right. I even retyped `BigInt` and it didn't click for me that that type is signed.

Since we're discussing this function, I suspect many instantiated `BigInt`s (in the wild) will be small and further, many will be bigger than one word. My point is that the range `2.1 billion.. 4.3 billion` is probably pretty uncommon, so it might make sense to just drop `u32` branch entirely. Alternatively it might make sense to try to cast to bigdigit for the unsigned branch (is there an easy way to do that?)

src/monty.rs
maxbla: Dang. well `for j in (0..=big_digit::BITS/n).map(|j| j * n)` seems to compile in rust `1.26`, but I'm not sure if that is cleaner. It would replace 3 lines.

src/biguint.rs
```diff
@@ -639,6 +656,21 @@ impl AddAssign<u64> for BigUint {
...
+
+    #[cfg(u64_digit)]
+    #[inline]
+    fn add_assign(&mut self, other: u64) {
+        if other != 0 {
+            if self.data.len() == 0 {
+                self.data.push(0);
+            }
+
+            let carry = __add2(&mut self.data, &[other as BigDigit]);
```
maxbla: The implementation of `__add2` has the comment `// Only for the Add impl:` above it. Since it no longer is only for add, maybe it can get some (short) documentation. It is still a private function of course.

It seems like `__add2` does something like "add assign ignoring carry, then return the carry". It might even make sense to rename `__add2`, but that it probably out of scope here.

src/biguint.rs
```diff
@@ -1912,30 +2078,66 @@ pub fn to_str_radix_reversed(u: &BigUint, radix: u32) -> Vec<u8> {
     res
 }
 
+/// Creates and initializes a `BigUint`.
+///
+/// The digits are in little-endian base matching `BigDigit`.
+///
+/// This is an internal `pub(crate)`-ish API only!
+#[inline]
+pub fn biguint_from_vec(digits: Vec<BigDigit>) -> BigUint {
```
maxbla: This function is `pub`, but it requires a `Vec<BigDigit>`, so it effectively can't be used externally. The `#[doc(hidden)]` attribute would hide it form public documentation. There technically could be a name conflict if someone did `use num_bigint::*`, but that seems unlikely.

src/bigint.rs
cuviper: The conversion checks should be relatively trivial, compared to the actual division. `Rem<i32>` even forwards to `Rem<u32>` with added sign handling, which is why it seems to make sense to deal with `u32` directly first. I don't feel strongly, but it feels like a micro-optimization to worry about this much, where the real win is just in dividing by a primitive.

Maybe you'd like to focus on this for a followup performance PR?

src/biguint.rs
cuviper: This is in the private `mod biguint` though, so it's truly inaccessible outside of the crate, as long as we don't accidentally re-export it elsewhere. It should be made `pub(crate)` whenever we raise our rustc minimum sufficiently though.

src/biguint.rs
cuviper: I think `__add2` predates `AddAssign`, so it was formerly just for `Add`. Then when we implemented `AddAssign`, it was convenient to forward `Add` to that.

I added some documentation to it.

src/bigint.rs
maxbla: Yeah, it would be silly to make this type of change without measuring. I am interested in doing some performance work, but I'm not exactly sure how to do it. It seems like num-bigint has used criterion.rs at some point? but I couldn't find it in the git log. I think criterion would probably be very helpful for optimizing.

src/biguint.rs
maxbla: It seems like half my comments are due to oversight on my part :/

cuviper:
OK, let's finally take the plunge in the upcoming 0.3 release! Thanks again @dignifiedquire for your initial work, and also @maxbla for helping with the review.

bors r+

PR #67, opened 2018-09-23
Update README.md alternatives section for apint

Robbepop:
The apint crate is stable as of Rust version 1.26.


cuviper:
Nice!

bors r+

PR #72, opened 2018-12-14
Fix `to_signed_bytes_*` for positive number with leading `0x80`

leodasvacas:
The trick for saving a byte on `0x800000..` only works for negative numbers. This caused +128 to be serialized as `0x80` which is -128. So pretty bad bug.


leodasvacas:
@cuviper Formatted it. Yhea, could be something to fuzz even.

cuviper:
I've got that round trip test written that also shows your failure.  I'll push it here in a moment.

Fuzzing num-bigint would be awesome across the board!

cuviper:
I published 0.2.2 -- thanks again!

leodasvacas:
@cuviper Thanks for the super fast turnaround on this one!

PR #77, opened 2019-01-09
Added Pow for Big types

mikelodder7:
Signed-off-by: Michael Lodder <redmike7@gmail.com>

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -840,24 +840,6 @@ fn powsign<T: Integer>(sign: Sign, other: &T) -> Sign {
     }
 }
 
-impl<'a> Pow<BigUint> for &'a BigInt {
-    type Output = BigInt;
-
-    #[inline]
-    fn pow(self, rhs: BigUint) -> BigInt {
-        BigInt::from_biguint(powsign(self.sign, &rhs), (&self.data).pow(rhs))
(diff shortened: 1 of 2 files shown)


cuviper:
Hi, thanks for the PR, and sorry for the late review.

I think we can avoid a little overhead in `exp` by using identities and `u64` as a limit.
- x<sup>0</sup> == 1 (including 0<sup>0</sup> -- https://github.com/rust-num/num-traits/pull/79)
- For e > 0, 0<sup>e</sup> == 0 and 1<sup>e</sup> == 1
- For e <= `u64::MAX`, we can cast it and use the `Pow<u64>` impl
    - Use a checked-cast, `ToPrimitive::to_u64`
    - The _least_ case of supporting that maximum is 2<sup>u64::MAX</sup>, which would take 2<sup>61</sup> = 2.3e18 bytes of memory = 2.3 exabytes!
    - No processor supports that much physical memory, even in theory:\
      https://en.wikipedia.org/wiki/64-bit_computing#Limits_of_processors
- Thus we can panic-overflow any e > `u64::MAX` (except for 0 or 1 base identities)

cuviper:
I rebased and implemented the simplification that I mentioned.

bors r+

PR #81, opened 2019-02-14
Add a quicker Rem

lcnr:
Right now all implementations of Rem return a `BigUint` and are somewhat slow.
Due to the fact that `a % b < b` is always true, we might want to change Rem to return a value of type `b` instead of a.

I have implemented a somewhat faster version of Rem and ran [some benchmarks](https://github.com/lcnr/quick-rem):

|xbits | old | new |old (bigger is better)
|---|----|----|--------------------
4 | 51,257 ns/iter (+/- 22,916)| 10,864 ns/iter (+/- 9,191)|4.72
6|67,351 ns/iter (+/- 44,639)| 23,170 ns/iter (+/- 19,247)|2.91
8 |132,859 ns/iter (+/- 54,852)| 75,033 ns/iter (+/- 6,360)|1.77
10 |399,424 ns/iter (+/- 75,839)| 346,401 ns/iter (+/- 2,146)|1.153
14 |5,957,626 ns/iter (+/- 394,632)| 5,986,485 ns/iter (+/- 1,919,245)|1.00
18 |96,009,267 ns/iter (+/- 12,585,847)| 94,161,577 ns/iter (+/- 15,044,440)|1.02

Relevant because of this [test for truncatable primes](https://github.com/lcnr/truncatable/blob/master/src/is_prime.rs) which is three times faster when using `quick_rem`instead of the current implementation.

### DIFF
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -1824,13 +1824,25 @@ pub fn to_str_radix_reversed(u: &BigUint, radix: u32) -> Vec<u8> {
 }
 
 impl BigUint {
+    /// quickly computes `self % div`
+    pub fn quick_rem(&self, div: u32) -> u32 {
+        let mut rem = 0u64;
+        for digit in self.data.iter().rev() {
+            rem = (rem << 32) + (*digit) as u64;
+            rem %= div as u64;
+        }
+
+        rem as u32
+    }
+
+
     /// Creates and initializes a `BigUint`.
     ///
     /// The digits are in little-endian base 2<sup>32</sup>.
     #[inline]
     pub fn new(digits: Vec<u32>) -> BigUint {
         BigUint { data: digits }.normalized()
-    }
+}
 
     /// Creates and initializes a `BigUint`.
     ///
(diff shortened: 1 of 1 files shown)


cuviper:
Before adding a new separate method, I think we can improve on the normal operator.

> Right now all implementations of Rem return a `BigUint` and are somewhat slow.

The output is one consideration, but I think the bigger issue for performance is that we're always computing the quotient and remainder together, just to return the latter.

https://github.com/rust-num/num-bigint/blob/eb003cc41b604658c885c052583ea9f453f9c910/src/biguint.rs#L1010-L1018

Currently we always need `self` by value (requiring a clone if we start with LHS as a reference), because `div_rem_digit` will write the quotient in place. I think we could change this to a simpler `rem_digit` which takes the LHS by reference, and doesn't need to build the quotient at all.

> Due to the fact that `a % b < b` is always true, we might want to change Rem to return a value of type `b` instead of a.

That's a reasonable possibility, but it would be a breaking change.

lcnr:
will do

PR #83, opened 2019-03-17
Use from() for lossless numerical type conversion

### LINKED ISSUE #66: Use from/into instead of as, to ensure type conversions are not lossy
There are quite a lot of places in the codebase (easy to see if you run `cargo clippy`) where, e.g. `val as u32` is done, instead of `u32::from(val)`. 

Clippy lint: https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#cast_lossless

mikong:
Resolves #66

### Note about clippy

As suggested in the issue, I used `cargo clippy` to find instances of casting using `as`. Running clippy reports a ton of warnings and some errors, so I used the following command to allow `clippy::all` but warn on `clippy::cast_lossless`, and output to file:

```
cargo clippy -- -A clippy::all -W clippy::cast_lossless > /path/to/output.txt 2>&1
```

### Unhandled cases

There were two cases reported by clippy which I did **not** handle. The first case is casting that involves macro invocation and the warning looks like the following:

```
warning: casting u8 to u32 may become silently lossy if types change
   --> src/macros.rs:220:41
    |
220 |                       $imp::$method(self, other as $promo)
    |  _________________________________________^
221 | |                 }
222 | |             }
223 | |
...   |
249 | |     (impl $imp:ident for $res:ty, $method:ident) => {
250 | |         promote_scalars!(impl $imp<u32> for $res, $method, u8, u16);
    | |______________________________________^ help: try: `u32::from(other)`
    | 
   ::: src/bigint.rs:964:1
    |
964 |   promote_all_scalars!(impl Add for BigInt, add);
    |   ----------------------------------------------- in this macro invocation
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cast_lossless
```

The second case was when clippy recommends `i64::from(n)` for example, but making the change causes a compilation error:

```
2351 |                 BigInt::from(i64::from(n))
     |                              ^^^^^^^^^ the trait `std::convert::From<isize>` is not implemented for `i64`
```

### Undetected

I also noticed that there are some instances of casting using `as` that are not reported by clippy. I'm not sure if this is a bug in clippy, or if there's a reason clippy didn't report them. For now, I also ignored these undetected instances.

### How the changes were made

For standard types, I simply followed clippy's suggestion, e.g. `u32::from(c)`.

For internal types such as `DoubleBigDigit`, clippy would suggest to use `u64::from()`, but I used `DoubleBigDigit::from()` instead.

To avoid any potential changes in behavior, I kept all parenthesis I encountered. In the following example, it looks safe to remove them, but I kept them all for consistency:

```
# previously:
*acc += (b as DoubleBigDigit) * (c as DoubleBigDigit);

# replaced with:
*acc += (DoubleBigDigit::from(b)) * (DoubleBigDigit::from(c));
```

Should I remove the parenthesis when it's clearly safe, but keep them when I'm not sure? Or maybe I should remove all of them, and let the tests catch any potential issue?

Please advise if I should try to fix the undetected instances. Also, I didn't review casting with `as` in the comments/documentation. Should I look into the comments too?

If I make these additional changes, let me know if I should squash all changes into one commit.

### DIFF
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -19,8 +19,8 @@ use big_digit::{self, BigDigit, DoubleBigDigit, SignedDoubleBigDigit};
 // Add with carry:
 #[inline]
 fn adc(a: BigDigit, b: BigDigit, acc: &mut DoubleBigDigit) -> BigDigit {
-    *acc += a as DoubleBigDigit;
-    *acc += b as DoubleBigDigit;
+    *acc += DoubleBigDigit::from(a);
+    *acc += DoubleBigDigit::from(b);
     let lo = *acc as BigDigit;
     *acc >>= big_digit::BITS;
     lo
@@ -29,8 +29,8 @@ fn adc(a: BigDigit, b: BigDigit, acc: &mut DoubleBigDigit) -> BigDigit {
 // Subtract with borrow:
 #[inline]
 fn sbb(a: BigDigit, b: BigDigit, acc: &mut SignedDoubleBigDigit) -> BigDigit {
-    *acc += a as SignedDoubleBigDigit;
-    *acc -= b as SignedDoubleBigDigit;
+    *acc += SignedDoubleBigDigit::from(a);
+    *acc -= SignedDoubleBigDigit::from(b);
     let lo = *acc as BigDigit;
     *acc >>= big_digit::BITS;
     lo
@@ -38,8 +38,8 @@ fn sbb(a: BigDigit, b: BigDigit, acc: &mut SignedDoubleBigDigit) -> BigDigit {
 
 #[inline]
 pub fn mac_with_carry(a: BigDigit, b: BigDigit, c: BigDigit, acc: &mut DoubleBigDigit) -> BigDigit {
-    *acc += a as DoubleBigDigit;
-    *acc += (b as DoubleBigDigit) * (c as DoubleBigDigit);
+    *acc += DoubleBigDigit::from(a);
+    *acc += DoubleBigDigit::from(b) * DoubleBigDigit::from(c);
     let lo = *acc as BigDigit;
     *acc >>= big_digit::BITS;
     lo
@@ -47,7 +47,7 @@ pub fn mac_with_carry(a: BigDigit, b: BigDigit, c: BigDigit, acc: &mut DoubleBig
 
 #[inline]
 pub fn mul_with_carry(a: BigDigit, b: BigDigit, acc: &mut DoubleBigDigit) -> BigDigit {
-    *acc += (a as DoubleBigDigit) * (b as DoubleBigDigit);
+    *acc += DoubleBigDigit::from(a) * DoubleBigDigit::from(b);
     let lo = *acc as BigDigit;
     *acc >>= big_digit::BITS;
     lo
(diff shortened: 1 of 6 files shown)


cuviper:
Hi, thanks for doing this.

I think for the macro-incurred errors, we should just allow the lints. Since we support older compilers, this will need to use the old style like `#[cfg_attr(feature = "cargo-clippy", allow(foo))]`, rather than the newer `#[allow(clippy::foo)]`.

For the parentheses, let's remove the obvious cases like `(DoubleBigDigit::from(x))`. If there's some other operator in the expression where precedence *might* be a question, we can keep that.

For undetected cases -- do you have an example? I wonder if they're actually lossy or not...

I don't mind multiple commits in a PR if they're building on each other. It's only when commits go back and forth, like cleaning up after unintended changes, then I prefer squashing that stuff.

mikong:
Noted on allowing the lints and that we should use the old style. I'll add a separate commit for this in this PR. I'll also address the parentheses soon after we decide what to do with the undetected cases.

Btw, the recent force-push in this PR is just to rebase my branch against the recent changes in master.

### Undetected Examples

I just checked src/bigint.rs and I found that almost all of the conversions were undetected. For example, here's the code for `negate_carry`:

https://github.com/rust-num/num-bigint/blob/447abda7ba1ab04708368baede925c1a0f8b6d2c/src/bigint.rs#L221-L226

The `as DoubleBigDigit` was the only conversion detected by clippy in the entire src/bigint.rs. The line with `as BigDigit` right after it was not detected. There are 50 more undetected conversions in this file. Here's another example:

https://github.com/rust-num/num-bigint/blob/447abda7ba1ab04708368baede925c1a0f8b6d2c/src/bigint.rs#L875-L881

Should I handle these instances not detected by clippy?

### rustfmt

Looks like rustfmt has been added to the CI. I suppose that's also another change I need to incorporate in this PR, correct? I can also add it as a separate commit in this PR so it's easy to track the changes.

cuviper:
Those "undetected" instances are *not* lossless conversions, so `From` won't work. The first `*acc as BigDigit` is a truncating conversion, and then the upper bits are shifted in `acc` for the next loop. In `i32_abs_as_u32`, this is a sign conversion which is also technically lossy, though safe in this context.

For `rustfmt`, yes I'll want that to pass. It looks like only a couple locations changed, so I don't mind either way whether it's a separate commit.

mikong:
Thanks for taking the time explaining why these aren't actually lossless conversions. I've updated my changes, with the first commit updated to remove the unnecessary parentheses in the conversions. The last commit is to address the rustfmt error.

The second commit is to allow the `cast_lossless` lint. I initially used just `#[cfg_attr(feature = "cargo-clippy", allow(cast_lossless))]`, but I get a warning:

```
warning: lint name `cast_lossless` is deprecated and may not have an effect in the future. Also `cfg_attr(cargo-clippy)` won't be necessary anymore
   --> src/macros.rs:218:60
    |
218 |                 #[cfg_attr(feature = "cargo-clippy", allow(cast_lossless))]
    |                                                            ^^^^^^^^^^^^^ help: change it to: `clippy::cast_lossless`
    | 
   ::: src/bigint.rs:976:1
    |
976 | promote_all_scalars!(impl Add for BigInt, add);
    | ----------------------------------------------- in this macro invocation
    |
    = note: #[warn(renamed_and_removed_lints)] on by default
```

The recommendation to change it to `#[cfg_attr(feature = "cargo-clippy", allow(clippy::cast_lossless))]` fixes the warning but causes a compilation error on Rust 1.15.0. So instead, I'm using the following which has been added to 3 places in src/macros.rs:

```
#[cfg_attr(feature = "cargo-clippy", allow(renamed_and_removed_lints))]
#[cfg_attr(feature = "cargo-clippy", allow(cast_lossless))]
```

cuviper:
Thanks, this looks good.

bors r+

PR #84, opened 2019-03-26
Add by value div_rem

lcnr:
Previously all calls to div/rem were using `algorithms::div_rem(u: &BigUint, d: &BigUint) -> (BigUint, BigUint)`. I renamed this method to `div_rem_ref` and added `algorithms::div_rem(u: BigUint, d: BigUint) -> (BigUint, BigUint)`. I also used this new method in `biguint.rs` when applicable. This reduces the amount of allocations needed.


cuviper:
Thanks, I rebased this and addressed the `set_one` TODO.

bors r+

PR #86, opened 2019-03-26
impl faster Rem<u32>

lcnr:
this implements/closes #81 

I ended up changing the macro names to be more reflective of what they are doing.


cuviper:
Thanks, this looks good - I just added a couple tiny tweaks. Sorry for such a tardy review!

bors r+

PR #93, opened 2019-05-02
Optimize BigUint::modpow for even

youknowone:
~This work is based on #77~

~This version only allow to take `&BigUint` as `exp` argument and doesn't manipulate `exp` not to copy the data of `exp`.~

### DIFF
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -2154,23 +2154,18 @@ impl BigUint {
     }
 }
 
-fn plain_modpow<'a, T>(base: &BigUint, exp_data: &Vec<T>, modulus: &BigUint) -> BigUint
-where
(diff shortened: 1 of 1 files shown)


cuviper:
While I can appreciate what you're trying to optimize, I think we need not bother -- https://github.com/rust-num/num-bigint/pull/77#issuecomment-488878023

Maybe you'd be interested to try this optimization in `modpow` instead? In that case it's actually reasonable to work with huge exponents.

youknowone:
Thanks for the comment. I will try  `modpow`

youknowone:
@cuviper I replaced the commits to modpow

cuviper:
Sorry for late review. Looks good -- I went ahead and pushed some tweaks of my own to this PR too.

bors r+

PR #97, opened 2019-07-13
Document ToBigInt behaviour

XAMPPRocky:
Adds documentation explaining why `ToBigInt` can return `None` in some cases, and when it can be ignored.


cuviper:
Thanks!

bors r+

PR #99, opened 2019-07-25
add quickcheck arbitrary and some quickcheck tests

### LINKED ISSUE #5: Add optional support for quickcheck::Arbitrary
Might be useful for testing, both for ourselves and for others using num-bigint.

(from rust-num/num#241)

maxbla:
Aims to fix #5. Still needs plenty of work, like supporting signed `BigInts`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,15 @@ features = ["std"]
 [dev-dependencies.serde_test]
 version = "1.0"
 
+[dev-dependencies.rand]
+version = "0.6"
+
+[dev-dependencies.quickcheck]
+version = "0.8"
+
+[dev-dependencies.quickcheck_macros]
+version = "0.8"
+
 [features]
 default = ["std"]
 i128 = ["num-integer/i128", "num-traits/i128"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,11 +56,10 @@ features = ["std"]
 [dev-dependencies.serde_test]
 version = "1.0"
 
-[dev-dependencies.rand]
-version = "0.6"
-
 [dev-dependencies.quickcheck]
-version = "0.8"
+version = "0.7"
+default-features = false
+features = []
 
 [features]
 default = ["std"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,9 +62,6 @@ version = "0.6"
 [dev-dependencies.quickcheck]
 version = "0.8"
 
-[dev-dependencies.quickcheck_macros]
-version = "0.8"
-
 [features]
 default = ["std"]
 i128 = ["num-integer/i128", "num-traits/i128"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 build = "build.rs"
 
 [package.metadata.docs.rs]
-features = ["std", "serde", "rand", "quick_check"]
+features = ["std", "serde", "rand", "quickcheck"]
 
 [[bench]]
 name = "bigint"
@@ -70,7 +70,6 @@ version = "1.0"
 default = ["std"]
 i128 = ["num-integer/i128", "num-traits/i128"]
 std = ["num-integer/std", "num-traits/std"]
-quick_check = ["quickcheck","quickcheck_macros"]
 
 [build-dependencies]
 autocfg = "0.1.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 build = "build.rs"
 
 [package.metadata.docs.rs]
-features = ["std", "serde", "rand"]
+features = ["std", "serde", "rand", "quickcheck"]
 
 [[bench]]
 name = "bigint"
@@ -53,13 +53,13 @@ version = "1.0"
 default-features = false
 features = ["std"]
 
+[dependencies.quickcheck]
+optional = true
+version = "0.8"
+default-features = false
+
 [dev-dependencies.serde_test]
 version = "1.0"
-
-[dev-dependencies.quickcheck]
-version = "0.7"
-default-features = false
-features = []
 
 [features]
 default = ["std"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,6 @@ version = "1.0"
 default = ["std"]
 i128 = ["num-integer/i128", "num-traits/i128"]
 std = ["num-integer/std", "num-traits/std"]
-quickcheck_testing = ["quickcheck", "quickcheck_macros"]
 
 [build-dependencies]
 autocfg = "0.1.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,7 @@ version = "1.0"
 default = ["std"]
 i128 = ["num-integer/i128", "num-traits/i128"]
 std = ["num-integer/std", "num-traits/std"]
+quickcheck_testing = ["quickcheck", "quickcheck_macros"]
 
 [build-dependencies]
 autocfg = "0.1.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 build = "build.rs"
 
 [package.metadata.docs.rs]
-features = ["std", "serde", "rand", "quickcheck"]
+features = ["std", "serde", "rand", "quick_check"]
 
 [[bench]]
 name = "bigint"
@@ -58,6 +58,11 @@ optional = true
 version = "0.8"
 default-features = false
 
+[dependencies.quickcheck_macros]
+optional = true
+version = "0.8"
+default-features = false
+
 [dev-dependencies.serde_test]
 version = "1.0"
 
@@ -65,6 +70,7 @@ version = "1.0"
 default = ["std"]
 i128 = ["num-integer/i128", "num-traits/i128"]
 std = ["num-integer/std", "num-traits/std"]
+quick_check = ["quickcheck","quickcheck_macros"]
 
 [build-dependencies]
 autocfg = "0.1.2"
--- a/ci/test_full.sh
+++ b/ci/test_full.sh
@@ -12,7 +12,7 @@ if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.31.0|1.26.0)$ ]]; then
   FEATURES="$FEATURES i128"
 fi
 if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.31.0)$ ]]; then
-  FEATURES="$FEATURES quickcheck quickcheck_testing"
+  FEATURES="$FEATURES quickcheck"
 fi
 
 # num-bigint should build and test everywhere.
--- a/ci/test_full.sh
+++ b/ci/test_full.sh
@@ -5,11 +5,14 @@ set -ex
 echo Testing num-bigint on rustc ${TRAVIS_RUST_VERSION}
 
 FEATURES="serde"
-if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.26.0|1.22.0)$ ]]; then
+if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.31.0|1.26.0|1.22.0)$ ]]; then
   FEATURES="$FEATURES rand"
 fi
-if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.26.0)$ ]]; then
+if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.31.0|1.26.0)$ ]]; then
   FEATURES="$FEATURES i128"
+fi
+if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.31.0)$ ]]; then
+  FEATURES="$FEATURES quickcheck"
 fi
 
 # num-bigint should build and test everywhere.
--- a/ci/test_full.sh
+++ b/ci/test_full.sh
@@ -12,7 +12,7 @@ if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.31.0|1.26.0)$ ]]; then
   FEATURES="$FEATURES i128"
 fi
 if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.31.0)$ ]]; then
-  FEATURES="$FEATURES quickcheck"
+  FEATURES="$FEATURES quickcheck quickcheck_macros"
 fi
 
 # num-bigint should build and test everywhere.
--- a/ci/test_full.sh
+++ b/ci/test_full.sh
@@ -12,7 +12,7 @@ if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.31.0|1.26.0)$ ]]; then
   FEATURES="$FEATURES i128"
 fi
 if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.31.0)$ ]]; then
-  FEATURES="$FEATURES quickcheck"
+  FEATURES="$FEATURES quick_check"
 fi
 
 # num-bigint should build and test everywhere.
--- a/ci/test_full.sh
+++ b/ci/test_full.sh
@@ -12,7 +12,7 @@ if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.31.0|1.26.0)$ ]]; then
   FEATURES="$FEATURES i128"
 fi
 if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.31.0)$ ]]; then
-  FEATURES="$FEATURES quick_check"
+  FEATURES="$FEATURES quickcheck"
 fi
 
 # num-bigint should build and test everywhere.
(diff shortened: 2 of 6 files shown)


maxbla:
The current version of Quickcheck is 0.8, which fails to build on our testing targets older than 1.31 because it uses the `dyn` keyword for traits. Quickcheck 0.7 does not do this. Quickcheck 0.7.x builds with all of our testing targets except 1.15. Quickcheck 0.6.x-0.7.x don't build for 1.15 because they use struct field shorthands. For the Rust 1.15 build to work, we would need to downgrade to Quickcheck 0.5.0. Should I do that? Does it make sense? I would love to keep using the highest version of Quickcheck possible, but if we really need to downgrade to Quickcheck 0.5, it wouldn't be the end of the world.

Cargo.toml
```diff
@@ -56,6 +56,11 @@ features = ["std"]
 [dev-dependencies.serde_test]
 version = "1.0"
 
+[dev-dependencies.quickcheck]
```
cuviper: Why is this a dev-dependency? As a regular dependency, it can be optional, and thus we can choose to support it only for sufficiently new compilers. It also needs to be that way for others to use it, per #5, rather than just `cfg(test)`.

So we could use 0.8 that way. The caveat is that it becomes a public dependency, so we can't bump semver on it without bumping semver ourselves too.

src/biguint.rs
```diff
@@ -2845,6 +2865,138 @@ fn get_radix_base(radix: u32) -> (BigDigit, usize) {
     }
 }
 
+#[cfg(test)]
+#[cfg(has_i128)]
+quickcheck! {
```
cuviper: Let's move all of these to a standalone file, like `tests/quickcheck.rs`, which can gate on the optional feature en masse. We also won't need `has_i128` if we're only supporting quickcheck on newer versions to begin with.

Cargo.toml
maxbla: I didn't realize how optional dependencies work. I'll switch that over now.

ci/test_full.sh
```diff
@@ -5,12 +5,15 @@ set -ex
 echo Testing num-bigint on rustc ${TRAVIS_RUST_VERSION}
 
 FEATURES="serde"
-if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.26.0|1.22.0)$ ]]; then
+if [[ "$TRAVIS_RUST_VERSION" =~ ^(nightly|beta|stable|1.31.0|1.26.0|1.22.0)$ ]]; then
```
maxbla: I figured that someone forgot to add 1.31 to the list of versions to test here. Please let me know if I got this wrong.

tests/quickcheck.rs
```diff
@@ -0,0 +1,326 @@
...
+            (a.clone() * a.clone() * a.clone()).cbrt() == a
+        }
+    }
+
+    quickcheck! {
+        fn quickcheck_unsigned_conversion(a:BigUint) -> bool {
+            let mut success = true;
+            for radix in 2..=36 {
+                let string = a.to_str_radix(radix);
+                success &= a == BigUint::from_str_radix(&string, radix).unwrap()
```
maxbla: I'll consider removing this function, as it takes ~5s to run on my computer

maxbla: maybe restrict radix to fewer values - 2,4,8,10,16,32?

maxbla:
I think this PR is ready for review.

src/biguint.rs
```diff
@@ -41,12 +41,32 @@ use UsizePromotion;
...
 pub struct BigUint {
     data: Vec<BigDigit>,
 }
 
+#[cfg(feature = "quickcheck")]
+impl Arbitrary for BigUint {
+    //Use arbitrary for Vec
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        let num = Vec::<u32>::arbitrary(g);
+        Self::from_slice(&num)
```
cuviper: Why a slice and not `Self::new(num)`? Let it take ownership.

src/biguint.rs
```diff
@@ -41,12 +41,32 @@ use UsizePromotion;
...
+    }
+    //Use the shrinker for Vec
+    fn shrink(&self) -> Box<Iterator<Item = Self>> {
+        Box::new(
+            self.data
+                .shrink()
+                .map(|x| BigUint::from_slice(x.as_slice())),
```
cuviper: Similarly, `map(Self::new)`.

tests/quickcheck.rs
```diff
@@ -0,0 +1,326 @@
...
+extern crate num_integer;
+extern crate num_traits;
+
+#[cfg(test)]
+#[cfg(feature = "quickcheck")]
+#[macro_use]
+pub extern crate quickcheck;
+
+#[cfg(test)]
+#[cfg(feature = "quickcheck")]
```
cuviper: We don't need `cfg(test)` in `tests/*.rs` files.

The feature test can be an inner attribute at the top to apply to this entire file, `#![cfg(...)]`. Then you don't need the additional `mod` at all.

tests/quickcheck.rs
cuviper: Maybe let the radix be another quickcheck-provided argument? I guess you can discard those outside of `2..=36`, or maybe there's a better way to deal with that...

tests/quickcheck.rs
```diff
@@ -0,0 +1,326 @@
...
+mod quickchecks {
+    use num_bigint::{BigInt, BigUint};
+    use num_traits::{Num, One, Pow, Zero};
+    use quickcheck::TestResult;
+
+    quickcheck! {
```
cuviper: This macro accepts repetitions, so you could write every `fn` within a single macro instance.

tests/quickcheck.rs
```diff
@@ -0,0 +1,326 @@
...
+        fn quickcheck_unsigned_eq_reflexive(a:BigUint) -> bool {
+            a == a
```
cuviper: I would be **shocked** if we ever failed a test like this, but as long as it runs quickly, I guess that's OK. Ditto for a lot of these tests that are testing broad mathematical properties.

tests/quickcheck.rs
```diff
@@ -0,0 +1,326 @@
...
+            match a.checked_add(b) {
+                None => TestResult::discard(),
+                Some(expected) => TestResult::from_bool(BigInt::from(expected) == actual)
+            }
+        }
+    }
+
+    quickcheck! {
+        fn quickcheck_unsigned_add_commutative(a: BigUint, b: BigUint) -> bool {
+            a.clone() + b.clone() == b + a
```
cuviper: A lot of these `clone()` calls. could do their operation by reference instead, `&a + &b == b + a`, as long as you trust that ref/value operators are relating correctly. Or maybe that's good as another angle for quickcheck to poke at anyway.

tests/quickcheck.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    quickcheck! {
+        fn quickcheck_signed_div_primitive(a: i128, b: i128) -> TestResult {
+            if b == 0 {
+                TestResult::discard()
+            } else {
+                TestResult::from_bool(BigInt::from(a/b) == BigInt::from(a) / BigInt::from(b))
+            }
+        }
+    }
```
cuviper: How about `Rem`, `Shl`, `Shr`, etc.? For example, we had a problem with `Shr` in #1, fixed to match primitives in #8, that would have been nice to discover sooner. I suspect quickcheck would have triggered it with ease. If you're curious, you could try breaking `shr_round_down` to return false and see what happens.

tests/quickcheck.rs
maxbla: Yeah, I shoved the equality tests in last minute with thinking somewhere along the lines "I rely on equality a lot. I should probably test it a little first." As for the broad mathematical properties, they seem a bit silly, I agree. They really aim to catch bugs that appear for large values, but are absent for smaller ones, otherwise I could just test every operation against primitives and catch every bug. I guess if using `u128`/`i128` primitives, there is probably full branch coverage (on the one thing I'm testing) even without the broad mathematical tests, since `BigInts` are based on `Vec<u32>`. I'm going to leave them in for now, but I think new tests should mostly test against (128 bit) primitives.

src/bigint.rs
```diff
@@ -114,6 +117,21 @@ pub struct BigInt {
...
 
+#[cfg(feature = "quickcheck")]
+impl Arbitrary for BigInt {
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        let num = BigUint::arbitrary(g);
+        let positive = bool::arbitrary(g);
+        let sign = if positive { Sign::Plus } else { Sign::Minus };
+        Self::from_biguint(sign, num)
+    }
+    fn shrink(&self) -> Box<Iterator<Item = Self>> {
```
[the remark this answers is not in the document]
maxbla: VS code's "prettier" add-on was doing this, which is weird because it doesn't claim to support Rust.

tests/quickcheck.rs
maxbla: Oh, I didn't realize. The way this is written is a holdover from the way I started (and the way I have previously used quickcheck) which is to have a function attribute i.e. `#[quickcheck]`. I switched from attributes to the macro in this crate to improve compatibility with old rust, but quickcheck didn't maintain compatibility with old Rust (the version requirements of the attribute are even tighter than that of the rest of quickcheck).

I prefer the quickcheck attribute on aesthetic grounds (less indentation), so I think I'll switch to that.

tests/quickcheck.rs
```diff
@@ -0,0 +1,283 @@
...
+    let mut success = true;
+    for radix in 2..=36 {
+        let string = a.to_str_radix(radix);
+        success &= a == BigInt::from_str_radix(&string, radix).unwrap()
+    }
+    success
+}
+
+#[quickcheck]
+fn quickcheck_unsigned_modpow(a: BigUint, exp: u8, modulus: BigUint) -> TestResult {
```
maxbla: Out of all the tests, this one actually took the longest time to run (~7s), and it isn't very useful, so I'm removing it.

Cargo.toml
```diff
@@ -53,13 +53,24 @@ version = "1.0"
...
+default-features = false
+
 [dev-dependencies.serde_test]
 version = "1.0"
 
 [features]
 default = ["std"]
 i128 = ["num-integer/i128", "num-traits/i128"]
 std = ["num-integer/std", "num-traits/std"]
+quick_check = ["quickcheck","quickcheck_macros"]
```
cuviper: While the attribute looks nice, I'm not keen on having this affect the public features. Let's just stick to the `quickcheck!` macro.

src/biguint.rs
```diff
@@ -41,12 +41,28 @@ use UsizePromotion;
...
 /// A big unsigned integer type.
 #[derive(Clone, Debug, Hash)]
 pub struct BigUint {
     data: Vec<BigDigit>,
 }
 
+#[cfg(feature = "quickcheck")]
+impl Arbitrary for BigUint {
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        //Use arbitrary from Vec
```
cuviper:
```suggestion
        // Use arbitrary from Vec
```

src/biguint.rs
```diff
@@ -41,12 +41,28 @@ use UsizePromotion;
...
+        Self::new(Vec::<u32>::arbitrary(g))
+    }
+
+    fn shrink(&self) -> Box<Iterator<Item = Self>> {
+        //Use shrinker from Vec
```
cuviper:
```suggestion
        // Use shrinker from Vec
```

Cargo.toml
maxbla: Yeah, I suspected you might not like this. Is there any better way to group optional dependencies? I was looking for something that would make it so you could select either `quickcheck` and `quickcheck_macros` or nothing, but found no way to accomplish that.

What do you mean by affect public features? If we don't re-export `quickcheck_macros`, it doesn't affect public features, right?

tests/quickcheck.rs
maxbla: There is a significant and annoying problem with testing shift operations (really, everything, but particularly shifts). `quickcheck` currently only generates numbers in `0..100` for unsigned types (I think `-100..100` for signed types). This makes for poor bit shift testing, because the upper bits of the number will always be zero. See https://github.com/BurntSushi/quickcheck/issues/190. I consider this issue blocking for bitshift quickchecks, as trivial tests pass when they would not if the range were the full primitive's range. Then if quickcheck were updated to check the whole range, it would break our tests (unless our tests were written to accommodate changing ranges, but this strategy seems risky). I may submit a PR to quickcheck for this.

For now I will investigate the comment on that issue:
> You can however change the range at which usize numbers are generated by creating a StdGen with a custom size and setting it on QuickCheck.

Cargo.toml
cuviper: By public features, I mean everything that would be valid for someone to add in their:

```toml
[dependencies.num-bigint]
version = "0.2"
features = [...]
```

All optional dependencies become named features, in addition to the explicit crate features, so here we would have all three choices available to the user: `quickcheck`, `quickcheck_macros`, and `quick_check`.

The most convenient solution for us would be if `quickcheck` had a feature to reexport the macros directly, just like `serde` does with its `derive` feature.

cuviper: Another downside is that we only need the macros for testing, but it will get forced in the build for all of our `quickcheck`-enabled users too. This is unfortunate because `syn` is pretty slow to build.

cuviper: Maybe just forget about `quick_check`? Users can enable `quickcheck` alone to get the implementations. Then we can enable `quickcheck_macros` when we run the tests, and gate on that `cfg` in the test file, but as a crate feature that will be useless for users.
(Enable this if you want to make your build slower! :wink: )

tests/quickcheck.rs
```diff
@@ -0,0 +1,305 @@
...
+        return TestResult::discard();
+    }
+    let string = a.to_str_radix(radix);
+    TestResult::from_bool(a == BigInt::from_str_radix(&string, radix).unwrap())
+}
+
+#[test]
+fn quicktest_shift() {
+    let gen = StdThreadGen::new(usize::max_value());
+    let mut qc = QuickCheck::with_gen(gen);
```
maxbla: I think it might be a good idea to do every test like this - to have a custom generator that generates full ranges for all types that aren't bigger than `usize`. Shame we still can't get the full range for `u128` this way, since `StdThreadGen::new` takes a `usize`.

tests/quickcheck.rs
maxbla: radix is now a quickcheck provided parameter. There's a slight problem in that ~65% of tests will be discarded. In this case I will leave as-is, but I could additionally hack on a `let radix = radix % 34 + 2`. That would make the test less readable (quickcheck might say it failed for radix = 100, but really radix is 100 % 34 + 2= 34) but the test would never be discarded.

tests/quickcheck.rs
maxbla: How much do you think I should trust that ref/value operators work properly? Isn't that implemented at a rust language level (i.e. there is no code in this crate to deal with it)? If that's the case, I don't feel the need to test it.

cuviper: It's not language level -- see `src/macros.rs` for what we use to implement value/reference combinations of operators. The forwarding is simple enough that I don't expect any bugs here, but you should also feel free to use it in testing (vs. clones).

tests/quickcheck.rs
cuviper: Should be mod 35 if we want to reach the maximum 36 -- `(0..35) + 2 = (2..37) = (2..=36)`. Or perhaps you could use `StdThreadGen` for a custom radix range?

maxbla: Whoops I got the formula wrong. I was just spit-balling though, that isn't in code.

Here's what the check looks like now:
```
    if radix > 36 || radix < 2 {
        return TestResult::discard();
    }
```
I think this is fine and I'd prefer to leave it this way, but in other situations, it might not be fine (like if the range were looking for was 0-3, as that is too low a percentage). Quickcheck aims for a certain number of tests to pass (i.e. not be discarded), but it has a maximum number of tests it will try.

> By default, if quickcheck can't find 100 valid tests after trying 10,000 times, then it will give up.

So there is no problem with the current implementation, which discards ~2/3 of tests.

I don't thing you could use a `StdThreadGen` to limit the range because quickcheck uses the same `StdThreadGen` for all variables in the function. So then we would end up with only really small numbers that are being converted.

tests/quickcheck.rs
maxbla: I tried to set an environment variable - `export QUICKCHECK_GENERATOR_SIZE=4294967295`, but this variable controls both the length of vectors generated and the maximum value of integers generated (so does `QuickCheck::with_gen()`), so setting it so high gives an allocation error `memory allocation of 16691263072 bytes failederror: process didn't exit successfully:` `...`.

 We could write our own code to determine the length of a vector generated with Arbitrary, but use the default Arbitrary for u32. This as an alternative to using the custom `StdThreadGen` in tests.

I really wish quickcheck's defaults were a better for our use case.

maxbla: I'm going to call this good for now. I ended up making a custom `StdThreadGen` for all the tests based on primitives i.e.
```rust
let gen = StdThreadGen::new(usize::max_value());
let mut qc = QuickCheck::with_gen(gen);
```
quickchecks taking `BigUint`/`BigInt` use the standard quickcheck gen (which isn't ideal, but I expect to find many bugs). This gen implementation can only check values where all `BigDigit`s have values 0-100 and there are 0-100 big digits. good enough.

tests/quickcheck.rs
maxbla: I think I'll use the reference operators. I expect it to test more code/branches. Two wrongs can't make a right 100 times in a row for random inputs. By which I mean that I don't have to trust both ref and value operators since ref operators are implemented in terms of value operators. If there is a bug in either the ref macros or the operator implementation I would probably catch it. Besides that, I expect the ref implementation to be faster, as it avoids needless copying.

cuviper:
I tried that shift test with intentionally broken `shr_round_down`, and it was indeed caught!

- always returning `false`:  
  `thread 'quicktest_shift' panicked at '[quickcheck] TEST FAILED. Arguments: (-1, 1)'`
- always returning `true`: 
  `thread 'quicktest_shift' panicked at '[quickcheck] TEST FAILED. Arguments: (1, 0)'`
- and of course, it passes with the proper `shr_round_down`

Thanks!

bors r+

PR #104, opened 2019-08-16
Add to_u32_digits to both BigInt and BigUint

Speedy37:
This add access to the u32 internal representation but because it returns a Vec of u32, doesn't force you to commit to this internal representation.

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -2777,6 +2777,24 @@ impl BigInt {
         (self.sign, self.data.to_bytes_le())
     }
 
+    /// Returns the sign and the u32 digits representation of the `BigInt` in little-endian order.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use num_bigint::{BigInt, Sign};
+    ///
+    /// assert_eq!(BigInt::from(-1125).to_u32_digits(), (Sign::Minus, vec![1125]));
+    /// assert_eq!(BigInt::from(4294967295u32).to_u32_digits(), (Sign::Plus, vec![4294967295]));
+    /// assert_eq!(BigInt::from(4294967296u64).to_u32_digits(), (Sign::Plus, vec![0, 1]));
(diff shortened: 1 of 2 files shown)


src/biguint.rs
```diff
@@ -2120,6 +2120,21 @@ impl BigUint {
...
     }
 
+    /// Returns the u32 digits representation of the `BigUint` in little-endian order.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use num_bigint::BigUint;
+    ///
+    /// let i = BigUint::parse_bytes(b"1125", 10).unwrap();
```
cuviper: `BigUint::from` would be infallible, no `unwrap` needed.
(might need to specify `1125u32` so it doesn't default to `i32`)

src/biguint.rs
```diff
@@ -2120,6 +2120,21 @@ impl BigUint {
...
+    /// assert_eq!(i.to_u32_digits(), vec![1125]);
```
cuviper: Can you include examples with multiple digits to demonstrate the order? Even just `i << 32` should make this easy to see.

src/bigint.rs
```diff
@@ -2777,6 +2777,21 @@ impl BigInt {
...
     }
 
+    /// Returns the sign and the u32 digits representation of the `BigInt` in little-endian order.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use num_bigint::{ToBigInt, Sign};
+    ///
+    /// let i = -1125.to_bigint().unwrap();
```
cuviper: Use `BigInt::from` to be infallible please.

src/bigint.rs
```diff
@@ -2777,6 +2777,21 @@ impl BigInt {
...
+    /// assert_eq!(i.to_u32_digits(), (Sign::Minus, vec![1125]));
```
cuviper: This too could use more examples.

cuviper:
Thanks!

bors r+

PR #105, opened 2019-08-17
Add PartialEq and PartialOrd between built in integer types and BigInt/BigUint

hansihe:
This has been one of the major annoyances for me when using num-bigint compared to something like rug.

### DIFF
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -739,6 +739,34 @@ pub fn biguint_shr(n: Cow<BigUint>, bits: usize) -> BigUint {
     BigUint::new(data)
 }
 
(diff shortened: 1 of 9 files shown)


cuviper:
On the surface this seems fine, although I wouldn't include the signed primitives with `BigUint`.

I'm concerned about what this does to type inference though, shown by the changes you had to make to existing tests. I think technically this is a *minor* breaking change according to Rust semver, which is allowable but deserves caution.

I think we'll probably have a semver bump in the near future for rand (#75), so it might be better to wait for that and document this as an upgrade hazard.

hansihe:
I removed the signed integer implementations for BigUint.

birkenfeld:
+1 for this patch.  I was wondering why `Add` etc. are implemented but not the comparison traits.

cuviper:
Included in #136.

PR #107, opened 2019-08-23
Fix lcm (0,0) panic

### LINKED ISSUE #103: division by zero when calculating lcm
`BigInt::zero().lcm(&BigInt::zero())` panics instead of returning the expected result of 0.

TheLetterTheta:
Fixes #103 
Added a simple check to the BigInt::lcm() in the case that both arguments are
zero. If this is the case, the method returns BigInt::zero().

Added test case to test/bigint.rs checking that lcm(0, 0) returns 0.

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -2188,7 +2188,11 @@ impl Integer for BigInt {
     /// Calculates the Lowest Common Multiple (LCM) of the number and `other`.
     #[inline]
     fn lcm(&self, other: &BigInt) -> BigInt {
-        BigInt::from_biguint(Plus, self.data.lcm(&other.data))
+        if self.is_zero() && other.is_zero() {
+            Self::zero()
+        } else {
+            BigInt::from_biguint(Plus, self.data.lcm(&other.data))
+        }
     }
 
     /// Deprecated, use `is_multiple_of` instead.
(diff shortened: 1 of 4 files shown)


src/bigint.rs
```diff
@@ -2188,7 +2188,11 @@ impl Integer for BigInt {
     /// Calculates the Lowest Common Multiple (LCM) of the number and `other`.
     #[inline]
     fn lcm(&self, other: &BigInt) -> BigInt {
-        BigInt::from_biguint(Plus, self.data.lcm(&other.data))
+        if self.is_zero() && other.is_zero() {
+            Self::zero()
+        } else {
+            BigInt::from_biguint(Plus, self.data.lcm(&other.data))
+        }
```
cuviper: Do we really need `BigInt` to check this condition? I would think that if `BigUint` does the right thing, `BigInt` should get it for free.

TheLetterTheta: You are correct. I thought the test had failed for `BigInt` when I added the test case.

cuviper: Looks like you added `BigUint` second, so I guess you did need this at first.

cuviper:
Thanks!

bors r+

PR #109, opened 2019-09-06
Document the "rand" feature.

alex-ozdemir:
When using this crate I had trouble figuring out why I couldn't use the `RandomBits` structure. Hopefully future users will be less confused!

Doc locations:
   * README
   * `RandomBits` struct
   * `RandBigInt` trait

### DIFF
--- a/src/bigrand.rs
+++ b/src/bigrand.rs
@@ -14,6 +14,9 @@ use bigint::{into_magnitude, magnitude};
 use integer::Integer;
 use traits::Zero;
 
+/// A trait for sampling random big integers.
+///
+/// The `rand` feature must be enabled to use this. See crate-level documentation for details.
 pub trait RandBigInt {
     /// Generate a random `BigUint` of the given bit size.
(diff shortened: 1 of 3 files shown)


src/bigrand.rs
```diff
@@ -14,6 +14,9 @@ use bigint::{into_magnitude, magnitude};
 use integer::Integer;
 use traits::Zero;
 
+/// A trait for sampling random big integers.
+///
+/// The `rand` feature must be enabled to use this. See crate-level documentation for details.
```
cuviper: Do you mean to reference `README.md`? The crate-level documentation would be at the top of `lib.rs`, which doesn't mention features at all. (But it maybe should!)

alex-ozdemir: Whoops! Good catch.

I'm not sure if there's an existing convention, but it seems reasonable to have features described in both places.

cuviper:
Thanks!

bors r+

PR #110, opened 2019-09-07
Bump dependencies

kpcyrd:
This bumps rand and quickcheck to the latest version. We're about to ship num-bigint with this patch applied.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
       before_script:
         - rustup target add $TARGET
       script:
-        - cargo build --verbose --target $TARGET --no-default-features --features "serde"
+        - cargo build --verbose --target $TARGET --no-default-features --features "serde rand"
     - name: "rustfmt"
       rust: 1.31.0
       before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
       before_script:
         - rustup target add $TARGET
       script:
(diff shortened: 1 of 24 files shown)


cuviper:
The code looks correct. I've also been playing with this locally, and we might want to tweak what `rand` features are pulled in, especially where they're only for dev-deps, but this is a minor point.

However, this is breaking change, since these are public dependencies -- you're changing public trait implementations to semver-incompatible versions.

> We're about to ship num-bigint with this patch applied.

Who is "we"? Are you shipping this as an internal detail, or is it something that will be publicly visible and different than the crates.io `num-bigint`?

kpcyrd:
oops, didn't pay enough attention when writing the text, "we" is referring to the debian rust team, the crate is used when building packages that depend on num-bigint. semver-incompatible reexports aren't much of a problem in that case.

programmerjake:
see also https://github.com/rust-num/num-bigint/issues/75

cuviper:
I took the liberty of rebasing this and stacking some more changes on top. The MSRV is now 1.31, and 128-bit integer support is assumed. MSRV with rand, quickcheck, and no_std are all higher, respectively 1.32, 1.34, and 1.36. We now support rand on no_std too, and I moved some of the rand and quickcheck tests to sub-crates to avoid "polluting" the main dependencies. I want to do a Rust 2018 conversion too, but I'll do that separately.

Thanks!

bors r+

cuviper:
Some 503 errors from crates.io -- let's try again.

bors retry

PR #113, opened 2019-10-06
Fix modpow for exp=1

youknowone:
Unfortunately, there was cases modulus never happened

### DIFF
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -2268,7 +2268,7 @@ fn plain_modpow(base: &BigUint, exp_data: &[BigDigit], modulus: &BigUint) -> Big
         Some(i) => i,
     };
 
-    let mut base = base.clone();
+    let mut base = base % modulus;
     for _ in 0..i {
         for _ in 0..big_digit::BITS {
             base = &base * &base % modulus;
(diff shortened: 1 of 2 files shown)


src/biguint.rs
```diff
@@ -2285,6 +2285,9 @@ fn plain_modpow(base: &BigUint, exp_data: &[BigDigit], modulus: &BigUint) -> Big
 
     let mut exp_iter = exp_data[i + 1..].iter();
     if exp_iter.len() == 0 && r.is_one() {
+        if &base > modulus {
+            base %= modulus;
+        }
```
cuviper: I think it would be better to normalize this above where we create a `base.clone()`, just use `base % modulus` instead. The implementation of that `%` operator already checks if the value is less than the modulus too.

tests/modpow.rs
```diff
@@ -81,6 +81,7 @@ mod biguint {
         check_modpow::<u32>(0, 15, 11, 0);
         check_modpow::<u32>(3, 7, 11, 9);
         check_modpow::<u32>(5, 117, 19, 1);
+        check_modpow::<u32>(20, 1, 2, 0);
```
cuviper: Please add a similar test with an odd modulus too, so that corner case is not missed. (But I think that implementation is already OK.)

youknowone: Additionally, I added api surface test for small values below

cuviper:
Thanks!

bors r+

PR #114, opened 2019-10-09
Fix BigInt::modpow behavior for negative base

youknowone:
For signed base, the sign was calculated in wrong way when exp is even


cuviper:
Thanks again! I rebased this to pick up #113, and I added a little more to the testing.

bors r+

youknowone:
@cuviper Is there a plan to release a new version with this patch?

cuviper:
Sorry about that -- it's now released in 0.2.4.

youknowone:
Thanks for the notice!

PR #115, opened 2019-10-21
Update rand to 0.7

mikelodder7:
Update code to work with the newer version of the `rand` crate.


cuviper:
Thanks, but this is already part of #110, and it's a breaking change as well.

PR #123, opened 2020-01-02
Add TryFrom impls

### LINKED ISSUE #120: Implement `TryFrom`/`TryInto`
Currently conversion between `BigUInt`, `BigInt` and primitives use dedicated traits `ToBigInt`, `ToBigUInt` and `{To|From}Primitive`, and not `Try{From|Into}` traits in `std`. While the functionality exists, this prevents using big integer types in a generic context that requires the latter. So, it would be preferable to implement them.

Some concerns:

* It might be better to integrate `ToBigInt` and the like directly with `Try{From|Into}`. But I think that will cause breakage downstream, especially with `{To|From}Primitive`. Still, if there is a way to do this with out breaking things, this seems preferable.

* Minimum supported rustc version. Since `Try{From|Into}` have been stable only since 1.34, I *think* naively implementing these would require bumping the rustc version number. Is there a way to avoid this? I'm not sure since I haven't dealt with minimum rustc versions before.

milesand:
Closes #120 

Currently contains:

*  `TryFrom` translations of existing conversion traits.
* A minimal Error type used for all `TryFrom` implementations.

Additional implemetations (e.g. `TryFrom<BigInt> for BigUint` which skips `clone()`) could be beneficial. Also, no new tests have been added since all new impls are basically thin adapters around existing, tested methods. It may be beneficial to add them regardless.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -192,9 +192,7 @@ pub struct TryFromBigIntError<T> {
 #[cfg(has_try_from)]
 impl<T> TryFromBigIntError<T> {
     fn new(original: T) -> Self {
-        TryFromBigIntError {
-            original
-        }
+        TryFromBigIntError { original }
     }
 
     fn __description(&self) -> &str {
@@ -203,7 +201,7 @@ impl<T> TryFromBigIntError<T> {
 
     /// Extract the original value, if available. The value will be available
     /// if the type before conversion was either [`BigInt`] or [`BigUint`].
-    /// 
+    ///
     /// [`BigInt`]: struct.BigInt.html
     /// [`BigUint`]: struct.BigUint.html
     pub fn into_original(self) -> T {
@@ -212,7 +210,10 @@ impl<T> TryFromBigIntError<T> {
 }
 
 #[cfg(all(feature = "std", has_try_from))]
-impl<T> std::error::Error for TryFromBigIntError<T> where T: fmt::Debug {
+impl<T> std::error::Error for TryFromBigIntError<T>
+where
+    T: fmt::Debug,
+{
     fn description(&self) -> &str {
         self.__description()
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -187,16 +187,23 @@ impl Error for ParseBigIntError {
 pub struct TryFromBigIntError(());
 
 #[cfg(has_try_from)]
+impl TryFromBigIntError {
+    fn __description(&self) -> &str {
+        "out of range conversion regarding big integer attempted"
+    }
+}
+
+#[cfg(all(feature = "std", has_try_from))]
 impl std::error::Error for TryFromBigIntError {
     fn description(&self) -> &str {
-        "out of range conversion regarding big integer attempted"
+        self.__description()
     }
 }
 
 #[cfg(has_try_from)]
 impl fmt::Display for TryFromBigIntError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        std::error::Error::description(self).fmt(f)
+        self.__description().fmt(f)
     }
 }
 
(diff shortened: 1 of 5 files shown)


cuviper:
FWIW, I rebased this to resolve the conflicts. It looks good!

> Additional implementations (e.g. `TryFrom<BigInt> for BigUint` which skips `clone()`) could be beneficial.

I'm still thinking about this before merging, because it would be a minor breaking change to add it later. For example, given `x: BigInt`, `x.try_into::<BigUint>()` will auto-ref right now, but will consume `x` if we add the additional implementations later. This could easily break code that still needed `x`.

I'm somewhat inclined to add by-value `TryFrom<BigInt>` and `TryFrom<BigUint>` in _all_ cases, even going into primitive integers, so they are useful in generic contexts like `T: TryInto<u32>` without forcing references.

src/lib.rs
```diff
@@ -182,6 +182,31 @@ impl Error for ParseBigIntError {
     }
 }
 
+#[cfg(has_try_from)]
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq, Hash)]
```
cuviper: I was just thinking that we might want to consider storing the original value in the error for `TryFrom<BigInt>`, but `Copy` would prevent that. We don't have to resolve that consideration right now, but let's remove `Copy` so we could do it later.

Also, `Default` and `Hash` seem questionable -- the standard `TryFromIntError` doesn't have those.

milesand: Storing the original value makes sense. However, currently the error's thrown at every point the conversion fails, so the original value can be one of multiple types: `BigInt`, `BigUint`, and maybe some primitives if we're storing those too.

Some possible solutions:

* Make the error type generic over original type: `TryFromBigIntError<T>`. I think this requires change now to be non-breaking.
* One error type for each original type; Actually a more verbose version of generic solution. Also requires change now.
* Store an enum of possible original types. Could be future-proofed by making the enum non-exhaustive. Has some runtime cost. I think keeping the current design ties us to this one, but I could be wrong.

So I believe this decision needs to be made before release; IMO going generic makes the most sense, but again I could be overlooking something.

milesand:
> I'm somewhat inclined to add by-value TryFrom<BigInt> and TryFrom<BigUint> in all cases

I pushed a relevant commit; This should cover most, if not all, by-value fallible conversion between primitives, `BigInt` and `BigUint`.

src/lib.rs
cuviper: > so the original value can be one of multiple types: `BigInt`, `BigUint`, and maybe some primitives if we're storing those too.

Also `&BigInt` and `&BigUint`, but I don't think there's much value in storing those, nor primitives. A lot of `dyn Error` APIs want `'static`, which makes references inconvenient.

So that would make 3 error categories: `TryFrom<BigInt>`, `TryFrom<BigUint>`, and the rest unsaved. I'm also leaning toward the generic solution, and I guess the unsaved can use `<()>`. Then the saved cases each need an extraction method, like `fn into_inner(self) -> BigInt`. I think those two can overlap the same method name.

> Store an enum of possible original types. Could be future-proofed by making the enum non-exhaustive.

I would want to keep an enum private, like parsing's `BigIntErrorKind` now, although it's also awkward to take different types out -- I guess it would need multiple methods returning `Option`.

milesand: I agree that the enum API would be rather awkward. I'll work on the generic design.

milesand: And it's done(406f192 and 0da7ae7). Specifically:

* The error is now generic over pre-conversion type
* Error sites modified to return correct error value
* `Error` impl now requires `T: Debug`
* Some docs on the error type added

The `Error` impl issue shouldn't matter much, since all our current `T`s implement `Debug`. It would be nicer if `T: !Debug` cases could be handled without making the whole Debug impl less useful, but I couldn't figure out how.

cuviper:
I added a couple tweaks for simplification, but otherwise it looks good, thanks!

bors r+

PR #133, opened 2020-01-25
Fix promotion of negative isize

### LINKED ISSUE #132: BigInt AddAssign<isize> with negative addend is off by 2^32 or 2^64
```rust
let mut x: BigInt = 0.into();
let y: isize = -1;
x += y;
assert_eq!(BigInt::from(-1), x); // fails
// x is actually 2^64-1
```
`x + y` gives the proper result (`-1`), and if `y` is `i32` or `i64` it works as expected, but `isize` somehow breaks things. When compiled for 64-bit the result is off by 2^64, and for 32-bit it is off by 2^32.

HactarCE:
This fixes #132.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -340,7 +340,7 @@ macro_rules! promote_signed_scalars {
 macro_rules! promote_signed_scalars_assign {
     (impl $imp:ident for $res:ty, $method:ident) => {
         promote_scalars_assign!(impl $imp<i32> for $res, $method, i8, i16);
-        promote_scalars_assign!(impl $imp<UsizePromotion> for $res, $method, isize);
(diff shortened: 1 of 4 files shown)


cuviper:
Ouch, thank you for finding this! Your change looks good.

It looks like that was a copy-paste typo all the way back from #41. The tests added then only cover native `BigInt` ops, not mixed with primitives. For the non-assign ops we have scalar tests that try conversions with all of the primitive integers, so we maybe should expand that for assign ops too. Is that something you could try? We should at least get a regression test of some sort.

The master branch is currently in preparation for 0.3, but I think I will open a branch and release a new 0.2.x for this fix too.

cuviper:
I just added more scalar tests myself. They did fail before your fix, and pass now.

bors r+

PR #136, opened 2020-02-25
impl PartialEq and PartialOrd with primitives

cuviper:
This rebases and extends #105 (cc @hansihe). As noted before, it may hurt type inference where callers use comparison operators that relied on only one possible `impl` before. We're on the way to 0.3 though, so even a relatively minor break like that is perfectly allowable.


hansihe:
:tada:

PR #143, opened 2020-04-04
Count big bits in u64 instead of usize

cuviper:
A 32-bit target _could_ create a giant `BigUint` over 2²⁹ bytes = 512MB,
not that this would be computationally useful, but that value would have
more than `usize::MAX` bits. To avoid that possibility of overflow, we
can better represent bit counts in `u64`.

For 64-bit targets, a `BigUint` over 2⁶¹ bytes is not realistic.

This changes the public API in the return values of `bits()` and
`trailing_zeros()`, as well as the `bit_size` parameter of `RandBigInt`
and `RandomBits`. Trying to randomize an absurd size will naturally
cause a capacity overflow, just as if you'd made a huge `Vec`.

### DIFF
--- a/benches/bigint.rs
+++ b/benches/bigint.rs
@@ -18,7 +18,7 @@ fn get_rng() -> StdRng {
     SeedableRng::from_seed(seed)
 }
 
-fn multiply_bench(b: &mut Bencher, xbits: usize, ybits: usize) {
+fn multiply_bench(b: &mut Bencher, xbits: u64, ybits: u64) {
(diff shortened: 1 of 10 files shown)


cuviper:
I'm not certain this is a good idea -- I'll think on it for a _bit_ before merging. Feedback welcome!

cuviper:
Ah, let's do it. If this makes it to 0.3 and folks complain, there's always 0.4 and beyond...

bors r+

PR #144, opened 2020-04-12
Add divide-by-zero panic messages

tech6hutch:
This brings it more in line with std's numbers, and lets the user actually know why it panicked.

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -3250,7 +3250,10 @@ impl BigInt {
             !exponent.is_negative(),
             "negative exponentiation is not supported!"
         );
-        assert!(!modulus.is_zero(), "divide by zero!");
+        assert!(
+            !modulus.is_zero(),
+            "attempt to calculate with zero modulus!"
+        );
 
         let result = self.data.modpow(&exponent.data, &modulus.data);
         if result.is_zero() {
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -1248,7 +1248,7 @@ impl Div<BigUint> for u32 {
     #[inline]
     fn div(self, other: BigUint) -> BigUint {
         match other.data.len() {
-            0 => panic!(),
+            0 => panic!("attempt to divide by zero"),
             1 => From::from(self as BigDigit / other.data[0]),
             _ => Zero::zero(),
         }
@@ -1280,7 +1280,7 @@ impl Div<BigUint> for u64 {
     #[inline]
     fn div(self, other: BigUint) -> BigUint {
         match other.data.len() {
-            0 => panic!(),
+            0 => panic!("attempt to divide by zero"),
             1 => From::from(self / u64::from(other.data[0])),
             2 => From::from(self / big_digit::to_doublebigdigit(other.data[1], other.data[0])),
             _ => Zero::zero(),
@@ -1291,7 +1291,7 @@ impl Div<BigUint> for u64 {
     #[inline]
     fn div(self, other: BigUint) -> BigUint {
         match other.data.len() {
-            0 => panic!(),
+            0 => panic!("attempt to divide by zero"),
             1 => From::from(self / other.data[0]),
             _ => Zero::zero(),
         }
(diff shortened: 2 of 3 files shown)


tech6hutch:
I hope I targeted the right branch, by the way. I was using the version on cargo, v0.2.6, when I noticed the lack of panic messages.

src/biguint.rs
```diff
@@ -1500,7 +1500,7 @@ impl Neg for BigUint {
 
     #[inline]
     fn neg(self) -> BigUint {
-        panic!()
+        panic!("unsigned values cannot be negated")
```
cuviper, line 1503: :open_mouth: I haven't noticed these nonsense `Neg` impls before -- that dates all the way back to `num`'s extraction from the standard library, before Rust 1.0. I'm inclined to remove them for `num-bigint` 0.3, but that's not for this PR...

src/biguint.rs
```diff
@@ -1420,7 +1420,7 @@ macro_rules! impl_rem_assign_scalar {
             fn rem_assign(&mut self, other: &BigUint) {
                 *self = match other.$to_scalar() {
                     None => *self,
-                    Some(0) => panic!(),
+                    Some(0) => panic!("attempt to calculate the remainder with a divisor of zero"),
```
cuviper, line 1423: Let's keep the various `Rem` impls consistent -- elsewhere it's using `div_rem`.

```suggestion
                    Some(0) => panic!("attempt to divide by zero"),
```

src/biguint.rs
cuviper: Here's more context on unsigned `Neg`: https://github.com/rust-lang/rust/pull/23945

cuviper:
> I hope I targeted the right branch, by the way. I was using the version on cargo, v0.2.6, when I noticed the lack of panic messages.

The master branch is good -- I hope to wrap up 0.3 "soon". :crossed_fingers: 

I've made a few bugfixes on 0.2 in the meantime, but I don't think I'd bother just for a diagnostic.

src/bigint.rs
```diff
@@ -3250,7 +3250,10 @@ impl BigInt {
             !exponent.is_negative(),
             "negative exponentiation is not supported!"
         );
-        assert!(!modulus.is_zero(), "divide by zero!");
+        assert!(
+            !modulus.is_zero(),
+            "attempt to calculate the remainder with a divisor of zero"
```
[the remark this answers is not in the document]
tech6hutch: Okay. I just copy-pasted the exact error message from rustc for doing int % 0.

src/biguint.rs
tech6hutch: I did wonder why you bothered impl-ing Neg at all for an unsigned type 😂

src/bigint.rs
```diff
@@ -3250,7 +3250,10 @@ impl BigInt {
             !exponent.is_negative(),
             "negative exponentiation is not supported!"
         );
-        assert!(!modulus.is_zero(), "divide by zero!");
+        assert!(
+            !modulus.is_zero(),
+            "attempt to calculate zero modulus!"
```
tech6hutch, line 3255: @cuviper is this message better?

src/biguint.rs
cuviper: FYI #145 removed it, so you'll need to drop your change to resolve that merge conflict.

src/bigint.rs
cuviper: Maybe "attempt to calculate with zero modulus"?

tech6hutch:
I did these changes from the GitHub editor, which is why there are so many commits. I assume you squash and merge anyway?

cuviper:
> I assume you squash and merge anyway?

I use a bot that does a plain merge, without squashing. I prefer to allow that PRs may have a series of meaningful commits, but in this case the separate commits are not helpful. So it would be great if you could squash this branch, or I can quickly do so myself.

tech6hutch:
> So it would be great if you could squash this branch, or I can quickly do so myself.

It'd be easier for you to do it, please. Thank you.

cuviper:
Thanks again!

bors r+

PR #158, opened 2020-06-28
Add iter_u32_digits , iter_u64_digits, to_u64_digits to both BigInt and BigUint

Speedy37:
This adds: `iter_u32_digits` , `iter_u64_digits`, `to_u64_digits` to both `BigInt` and `BigUint`

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -3007,8 +3007,8 @@ impl BigInt {
         (self.sign, self.data.to_u64_digits())
     }
 
-    /// Returns the `u32` digits representation of the `BigUint` ordered least significant digit
-    /// first.
+    /// Returns an iterator of `u64` digits representation of the `BigUint` ordered least
+    /// significant digit first.
     ///
     /// # Examples
     ///
@@ -3026,8 +3026,8 @@ impl BigInt {
         self.data.iter_u32_digits()
     }
 
-    /// Returns the `u64` digits representation of the `BigUint` ordered least significant digit
-    /// first.
+    /// Returns an iterator of `u64` digits representation of the `BigUint` ordered least
+    /// significant digit first.
     ///
     /// # Examples
     ///
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -31,7 +31,7 @@ use self::Sign::{Minus, NoSign, Plus};
 use crate::big_digit::{self, BigDigit, DoubleBigDigit};
 use crate::biguint;
 use crate::biguint::to_str_radix_reversed;
-use crate::biguint::{BigUint, IntDigits};
+use crate::biguint::{BigUint, IntDigits, IterU32Digits, IterU64Digits};
 use crate::ParseBigIntError;
 #[cfg(has_try_from)]
 use crate::TryFromBigIntError;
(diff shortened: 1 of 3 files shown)


Speedy37:
I'm wondering if adding `iter_u32_digits`/`iter_u64_digits` to `BigInt` is a good idea.
I guess users can use (`magnitude().iter_u32_digits()`) if they needs too.

src/bigint.rs
```diff
@@ -2988,6 +2988,63 @@ impl BigInt {
...
+    /// use num_bigint::BigInt;
+    ///
+    /// assert_eq!(BigInt::from(-1125).iter_u32_digits().collect::<Vec<u32>>(), vec![1125]);
+    /// assert_eq!(BigInt::from(4294967295u32).iter_u32_digits().collect::<Vec<u32>>(), vec![4294967295]);
+    /// assert_eq!(BigInt::from(4294967296u64).iter_u32_digits().collect::<Vec<u32>>(), vec![0, 1]);
+    /// assert_eq!(BigInt::from(-112500000000i64).iter_u32_digits().collect::<Vec<u32>>(), vec![830850304, 26]);
+    /// assert_eq!(BigInt::from(112500000000i64).iter_u32_digits().collect::<Vec<u32>>(), vec![830850304, 26]);
+    /// ```
+    #[inline]
+    pub fn iter_u32_digits<'a>(&'a self) -> IterU32Digits<'a> {
```
cuviper, line 3025: I think it's more idiomatic to omit lifetimes, but leave a marker:
```suggestion
    pub fn iter_u32_digits(&self) -> IterU32Digits<'_> {
```

src/biguint.rs
```diff
@@ -2255,6 +2266,200 @@ pub(crate) fn to_str_radix_reversed(u: &BigUint, radix: u32) -> Vec<u8> {
...
+            .unwrap_or(false);
+        IterU32Digits {
+            data,
+            next_is_lo: true,
+            last_hi_is_zero,
+        }
+    }
+}
+#[cfg(u64_digit)]
+impl<'a> Iterator for IterU32Digits<'a> {
```
cuviper, line 2293: Similarly on impls, the lifetime can be omitted when it's not otherwise named.
```suggestion
impl Iterator for IterU32Digits<'_> {
```

src/biguint.rs
```diff
@@ -2255,6 +2266,200 @@ pub(crate) fn to_str_radix_reversed(u: &BigUint, radix: u32) -> Vec<u8> {
...
+    }
+
+    fn count(self) -> usize {
+        self.len()
+    }
+}
+#[cfg(u64_digit)]
+impl<'a> ExactSizeIterator for IterU32Digits<'a> {
+    fn len(&self) -> usize {
+        self.data.len() * 2 - usize::from(self.last_hi_is_zero)
```
cuviper, line 2338: Doesn't this need to count whether `next_is_lo`?

src/biguint.rs
```diff
@@ -2255,6 +2266,200 @@ pub(crate) fn to_str_radix_reversed(u: &BigUint, radix: u32) -> Vec<u8> {
...
+impl<'a> IterU32Digits<'a> {
+    fn new(data: &'a [u32]) -> Self {
+        Self { it: data.iter() }
+    }
+}
+#[cfg(not(u64_digit))]
+impl<'a> Iterator for IterU32Digits<'a> {
+    type Item = u32;
+    fn next(&mut self) -> Option<u32> {
+        self.it.next().map(|&t| t)
```
cuviper, line 2356: I suggest `cloned()` instead of these maps -- which can be `copied()` when we raise MSRV.
```suggestion
        self.it.next().cloned()
```

src/biguint.rs
```diff
@@ -2516,30 +2714,61 @@ impl BigUint {
     /// ```
     #[inline]
     pub fn to_u32_digits(&self) -> Vec<u32> {
-        let mut digits = Vec::new();
+        self.iter_u32_digits().collect()
+    }
 
-        #[cfg(not(u64_digit))]
-        digits.clone_from(&self.data);
+    /// Returns the `u32` digits representation of the `BigUint` ordered least significant digit
```
cuviper, line 2720:
```suggestion
    /// Returns the `u64` digits representation of the `BigUint` ordered least significant digit
```

src/biguint.rs
```diff
@@ -2516,30 +2714,61 @@ impl BigUint {
...
-                    digits.push(x as u32);
-                    digits.push((x >> 32) as u32);
-                }
-                digits.push(last_lo);
-                if last_hi != 0 {
-                    digits.push(last_hi);
-                }
-            }
-        }
+    /// Returns the `u32` digits representation of the `BigUint` ordered least significant digit
```
cuviper, line 2738: I suggest either "Returns an iterator of ..." or "Iterates ..."

src/biguint.rs
```diff
@@ -2516,30 +2714,61 @@ impl BigUint {
...
+    /// assert_eq!(BigUint::from(4294967296u64).iter_u32_digits().collect::<Vec<u32>>(), vec![0, 1]);
+    /// assert_eq!(BigUint::from(112500000000u64).iter_u32_digits().collect::<Vec<u32>>(), vec![830850304, 26]);
+    /// ```
+    #[inline]
+    pub fn iter_u32_digits<'a>(&'a self) -> IterU32Digits<'a> {
+        IterU32Digits::new(self.data.as_slice())
+    }
 
-        digits
+    /// Returns the `u32` digits representation of the `BigUint` ordered least significant digit
```
cuviper, line 2756: Same comment about returns/iterates, plus this is `u64`.

src/bigint.rs
```diff
@@ -2988,6 +2988,63 @@ impl BigInt {
         (self.sign, self.data.to_u32_digits())
     }
 
+    /// Returns the sign and the `u32` digits representation of the `BigInt` ordered least
```
cuviper, line 2991:
```suggestion
    /// Returns the sign and the `u64` digits representation of the `BigInt` ordered least
```

src/bigint.rs
```diff
@@ -2988,6 +2988,63 @@ impl BigInt {
...
+    /// assert_eq!(BigInt::from(4294967296u64).iter_u32_digits().collect::<Vec<u32>>(), vec![0, 1]);
+    /// assert_eq!(BigInt::from(-112500000000i64).iter_u32_digits().collect::<Vec<u32>>(), vec![830850304, 26]);
+    /// assert_eq!(BigInt::from(112500000000i64).iter_u32_digits().collect::<Vec<u32>>(), vec![830850304, 26]);
+    /// ```
+    #[inline]
+    pub fn iter_u32_digits<'a>(&'a self) -> IterU32Digits<'a> {
+        self.data.iter_u32_digits()
+    }
+
+    /// Returns the `u32` digits representation of the `BigUint` ordered least significant digit
```
cuviper, line 3029:
```suggestion
    /// Returns the `u64` digits representation of the `BigUint` ordered least significant digit
```

src/biguint.rs
Speedy37: Yeah, I will add tests for this and fix it.

Speedy37:
Should `iter_u32_digits` and `iter_u64_digits` on `BigInt` also returns the `(Sign, Iterator)` couple to match other similar method return types?

src/bigint.rs
cuviper: Sorry I wasn't clear, but I meant for this change request to be extrapolated to the other similar methods too.

src/biguint.rs
cuviper: This too, there are several other impls like it that can omit the lifetime.

src/bigint.rs
Speedy37: Sorry, I should have checked.

cuviper:
I think it's OK that these methods return just the iterator, because that makes chaining easier, and `fn sign` can be used separately.

src/bigint.rs
```diff
@@ -2988,6 +2988,63 @@ impl BigInt {
...
+    /// assert_eq!(BigInt::from(4294967296u64).to_u64_digits(), (Sign::Plus, vec![4294967296]));
+    /// assert_eq!(BigInt::from(-112500000000i64).to_u64_digits(), (Sign::Minus, vec![112500000000]));
+    /// assert_eq!(BigInt::from(112500000000i64).to_u64_digits(), (Sign::Plus, vec![112500000000]));
+    /// ```
+    #[inline]
+    pub fn to_u64_digits(&self) -> (Sign, Vec<u64>) {
+        (self.sign, self.data.to_u64_digits())
+    }
+
+    /// Returns an iterator of `u64` digits representation of the `BigUint` ordered least
```
cuviper, line 3010:
```suggestion
    /// Returns an iterator of `u32` digits representation of the `BigUint` ordered least
```

Speedy37:
I fixed the latest set of requested changes.

cuviper:
Can you add `#[inline]` on all of the iterator methods? I tried myself, and it makes a huge difference to the benchmarks:

```
 name             base.x86_64 ns/iter  inline.x86_64 ns/iter  diff ns/iter   diff %  speedup
 iter_u32_digits  130                  62                              -68  -52.31%   x 2.10
 iter_u64_digits  64                   8                               -56  -87.50%   x 8.00
 to_u32_digits    152                  78                              -74  -48.68%   x 1.95
 to_u64_digits    85                   32                              -53  -62.35%   x 2.66
```

```
 name             base.i686 ns/iter  inline.i686 ns/iter  diff ns/iter   diff %  speedup
 iter_u32_digits  136                14                           -122  -89.71%   x 9.71
 iter_u64_digits  111                42                            -69  -62.16%   x 2.64
 to_u32_digits    154                62                            -92  -59.74%   x 2.48
 to_u64_digits    127                68                            -59  -46.46%   x 1.87
```

(I was going to just do it, but it seems your branch isn't open to maintainer pushes.)

Speedy37:
Done

cuviper (changes requested):
One more bikesheddy thought -- maybe we should drop the `iter_`/`Iter` prefixes? e.g. `u32_digits -> U32Digits`. I was considering the `str` iterators for prior art with `bytes() -> Bytes` and `chars() -> Chars`, or the maps have `keys -> Keys`, etc. I think it's uncommon to use `iter`/`Iter` in the names unless that's all there is to it (`Iter`/`IterMut`/`IntoIter`).

cuviper:
Completed in #192.

PR #159, opened 2020-07-18
Update bigint modpow documentation

mishazharov:
This commit corrects the range for the case that `modulus < 0`. I know it's a small inconsequential change but it's also an easy fix, so I figured why not.


mishazharov:
Never mind! The documentation is correct (Since `modulus < 0`, `(modulus, 0]` is actually the right range)

Issue #161, opened 2020-08-23
Infinite not being used


notoria:
Example:
```Rust
extern crate num;

use num::{BigInt, BigRational, ToPrimitive};

fn main() {
    let n = BigInt::from(-2);
    let n = n.pow(4001);
    println!("{:?}", n.to_f64());
    let r = BigRational::from(n);
    println!("{:?}", r.to_f64());
}
```
Result:
```
None
Some(-inf)
```
I was expecting some minus infinite and got none. A workaround is to convert to `BigRational` but it is expensive, a clone is required.

cuviper:
You can also see `ToPrimitive` sticking to finite values when narrowing, e.g. `std::f64::MAX.to_f32()` returns `None`. It's been that way ever since it was first created in rust-lang/rust#9250 (specifically [here](https://github.com/rust-lang/rust/blob/cb240197441723fed70e4076d9eba460d8f209ba/src/libstd/num/num.rs#L576-L591)), and it doesn't appear that this point was debated.

At that time, bigints only directly supported converting to 64-bit integers, and floats would convert from there by the trait default, so larger values would just be `None`. Then rust-num/num#145 added explicit conversions from bigints to floats, noting that `None` was used instead of `Some(INFINITY)`, but we didn't debate that point.

So this is actually inconsistent on `BigRational`'s part, which is relatively recent: rust-num/num-rational#52.

That's not to insist that we shouldn't change, but we should be careful and consistent about it.

cuviper:
Another inconsistency: `u128::MAX.to_f32()` returns `Some(INFINITY)`.

notoria:
Understood, thank you for clarifying the situation, closing.

cuviper:
I was starting to think we _should_ change -- do you think not?

notoria:
To make it consistent? Yes. Would it be for finite value or would infinite value be included?

cuviper:
I think we should include infinities -- if nothing else, the fact that they can be positive or negative is better than `None`.

notoria:
Agree. Reopening to track the issue.

PR #163, opened 2020-08-25
Extended the result of to_f32/to_f64 with infinity

notoria:


notoria:
This PR attempts to solve #161.

cuviper:
Thanks!

bors r+

ltratt:
This isn't a complaint on my part, but this change broke some of our code because we relied on "numbers that are too big to be an f32/f64" returning `None`. I wonder if, perhaps, this might want to be a major release (0.4) rather than a 0.3 point release? FWIW, I think both the original behaviour and the behaviour in this PR are reasonable so I'm not arguing for rolling back the behaviour.

cuviper:
@ltratt Sorry for the trouble! I debated about the change in behavior myself, but I decided that the sign preservation makes this better than `None`. It's already published in 0.3.1, so I don't think we should flip back without a stronger complaint, but I do appreciate your feedback!

PR #164, opened 2020-08-29
Add zero constants

mpdn:
This adds `ZERO` as constants for `BigInt` and `BigUint`. Having a constant is more useful than the `zero` function, as the lifetime of the const will be `'static` whereas the function will create a new zero on the stack.

### DIFF
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -501,7 +501,7 @@ impl_shift! { i8, i16, i32, i64, i128, isize }
 impl Zero for BigUint {
     #[inline]
     fn zero() -> BigUint {
-        BigUint { data: Vec::new() }
+        BigUint::ZERO.clone()
     }
 
     #[inline]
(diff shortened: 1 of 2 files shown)


mpdn:
This requires Rust 1.39.0.

src/biguint.rs
```diff
@@ -2264,6 +2264,11 @@ pub(crate) fn biguint_from_vec(digits: Vec<BigDigit>) -> BigUint {
 }
 
 impl BigUint {
+    /// The value 0 as a `BigUint`
+    pub const ZERO: BigUint = BigUint {
```
cuviper, line 2268: For MSRV, these new constants should be guarded by a `cfg`. You can use `autocfg::emit_constant_cfg` testing something like `Vec::<u32>::new()` in the build script. (Ditto on `BigInt`.)

src/biguint.rs
```diff
@@ -501,7 +501,7 @@ impl_shift! { i8, i16, i32, i64, i128, isize }
 impl Zero for BigUint {
     #[inline]
     fn zero() -> BigUint {
-        BigUint { data: Vec::new() }
+        BigUint::ZERO.clone()
```
cuviper, line 504: There's no need to `clone()` here. But since the `const` needs to be conditional anyway, let's just leave the original implementation as-is.  (Ditto on `BigInt`.)

cuviper:
If we do an inherent `pub const fn zero()`, there can also be an inherent `pub fn zero()` for older rustc, so we always have some method available without importing `Zero` -- just conditionally `const`. Maybe non-const `one()` should be added in tandem.

I don't want to get into mirroring _all_ of the trait methods as inherent methods, but constructors are particularly useful.

PR #166, opened 2020-09-11
Support arbitrary::Arbitrary

e00E:
num-bigint already supports quickcheck::Arbitrary which works roughly
like arbitrary::Arbitrary. arbitrary::Arbitrary used by cargo-fuzz and
makes it easier to use num-bigint BigInt and BigUint in projects that
want to fuzz that way.

https://github.com/rust-fuzz/arbitrary
https://github.com/rust-fuzz/cargo-fuzz

I have included an example fuzz target but I'm not sure whether we want to merge it.

```
cargo +nightly install cargo-fuzz
cargo +nightly fuzz run fuzz_target_1
```


cuviper:
It's unfortunate that there are similar but separate `Arbitrary` traits, but I guess they have different models for generating data.

I'm OK with this, but can you add something to `ci/test_full.sh` to at least build with that feature?

e00E:
I've added building the arbitrary feature to CI. Before merging we should still delete the `fuzz` folder. This was meant for reviewers but I don't think it makes sense in the repository.

cuviper:
> Before merging we should still delete the `fuzz` folder. This was meant for reviewers but I don't think it makes sense in the repository.

Ok, please do, and then I think we're good to go.

PR #171, opened 2020-10-24
Shrink overallocated capacity if a number gets a lot smaller.

tczajka:
If a number gets below 1/4 of capacity, shrink_to_fit. Since Vec grows by 2, this amortizes well
for any sequence of sizes.

### DIFF
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -2617,6 +2617,9 @@ impl BigUint {
         while let Some(&0) = self.data.last() {
             self.data.pop();
         }
+        if self.data.len() < self.data.capacity() / 4 {
+            self.data.shrink_to_fit();
+        }
     }
 
     /// Returns a normalized `BigUint`.


cuviper:
Do you have any real-world data supporting this? I'm hoping to see both memory use and performance impact. The performance is _probably_ fine, since it's a simple comparison of local fields, but it would be good to have measurements.

tczajka:
This is primarily motivated by my desire to have a theoretical upper bound worst case guarantee, rather than a gain on average.

If I am doing some huge calculation, and have n numbers of m bits, it is important for me to have an upper bound that says I'll be using O(nm) bits of allocated memory. Otherwise, using the library for such calculations would give me pause because I would never know if memory use won't blow up, have to trust a heuristic when it could be a guarantee.

Most benchmarks show no impact, some benchmarks show ~5% degradation (in CPU).

I suspect the impact is not because of the check, but because when we actually do shrink, there is no extra space left, and sometimes having extra space space (even one digit) helps in later operations (e.g. division currently always starts by shifting left by a few bits, which often needs one extra digit).

So using the `shrink_to` API (instead of `shrink_to_fit`) would probably help.

Similarly, when a number is first allocated, I speculate that adding a few words of extra capacity might help performance on benchmarks.

cuviper:
OK, I think that's reasonable, and we can keep an eye on using `shrink_to` and `with_capacity` for headroom later. I'm also not confident that all shrinking ops are covered by `normalize`, but there's no public promise here, so that seems ok for now.

bors r+

cuviper:
bors retry

Issue #174, opened 2020-11-08
BigUint should have count_ones() and similar methods


BartMassey:
I needed `BigUint::count_ones()` for a project, so I implemented it in terms of `to_u32_digits()`.

```
fn count_ones(b: BigUint) -> u32 {
    b
        .to_u32_digits()
        .into_iter()
	.map(|d| d.count_ones())
        .sum()
}
```

But now I'm relying on the optimizer to clean up my code, plus the implementation should probably return `u64` because it's pretty easily possible to have more than 4 billion ones in a `BigUint` so it's buggy. :-)

I'd be happy to put together some kind of pull request if there's some interest in one.

cuviper:
The iterators in #158 will at least let you do that without creating a temporary `Vec`, but there will still be some translation from the internal digit size, which is extraneous for your task. Internally iterating `self.data` would be ideal, so yes, I'm open to a pull request for `BigUint::count_ones(&self) -> u64`.

A harder question is whether we should do anything for `BigInt`, since negative numbers conceptually have an infinite prefix of 1s in twos-complement (though we store sign+magnitude). We could count ones in the magnitude instead, but I think it would be unfortunate to have such different semantics from `iN::count_ones`. I suppose folks can use `bigint.magnitude().count_ones()` if that's what they want.

RE: "and similar methods": For `count_zeros`, I guess we could define that as only counting less than the MSB 1. For `leading_zeros` we'd have the same problem of the infinite prefix, but `BigUint::leading_ones` should be okay. We do have `trailing_zeros` already, and fortunately that is even consistent between twos-complement and the magnitude. I think that would be mostly true for `trailing_ones` too, but `BigInt == -1` would need to be an exception.

tczajka:
It's not true for `trailing_ones`, e.g. `(-3).trailing_ones() == 1` while `3.trailing_ones() == 2`.

cuviper:
@tczajka Ah, you're right, of course. We can still do `BigUint::trailing_ones` easily, but we'd have to take care with `BigInt`.

BartMassey:
I think leaving `BigInt` out for now is probably the path of least resistance. Given that `BigInt` is sign-magnitude the users can do what they want after the essentially-free conversion to `BigUint`.

Should `count_ones()` return `u64` or `usize`? I could make the case either way: neither is ideal. We could make it return `BigUint` but that seems needlessly silly for most applications…

I'll try to get a PR together. Thanks much!

[...]

PR #175, opened 2020-11-10
added BigUint::count_ones(), trailing_ones()

BartMassey:
Added `BigUint::count_ones()`, `BigUint::trailing_ones()`. Also minimal tests for these and `BigUint::trailing_zeros()`.

Closes Issue #174

### DIFF
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -2689,6 +2689,24 @@ impl BigUint {
         let zeros: u64 = self.data[i].trailing_zeros().into();
         Some(i as u64 * u64::from(big_digit::BITS) + zeros)
     }
+
+    /// Returns the number of least-significant bits that are ones.
+    pub fn trailing_ones(&self) -> u64 {
+        if let Some(i) = self.data.iter().position(|&digit| !digit != 0) {
+            // XXX u64::trailing_ones() introduced in Rust 1.46,
+            // but we need to be compatible further back.
+            // Thanks to cuviper for this workaround.
+            let ones: u64 = (!self.data[i]).trailing_zeros().into();
+            i as u64 * u64::from(big_digit::BITS) + ones
+        } else {
+            self.data.len() as u64 * u64::from(big_digit::BITS)
+        }
+    }
+
+    /// Returns the number of one bits.
+    pub fn count_ones(&self) -> u64 {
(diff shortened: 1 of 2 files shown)


src/biguint.rs
```diff
@@ -2689,6 +2689,34 @@ impl BigUint {
         let zeros: u64 = self.data[i].trailing_zeros().into();
         Some(i as u64 * u64::from(big_digit::BITS) + zeros)
     }
+
+    /// Returns the number of significant bits: bit position
+    /// of leftmost 1-bit + 1, or 0 for no 1-bits (number is 0).
+    pub fn significant_bits(&self) -> u64 {
```
cuviper, line 2695: I think this is already covered by `pub fn bits(&self) -> u64`.

src/biguint.rs
```diff
@@ -2689,6 +2689,34 @@ impl BigUint {
...
+        } else {
+            0
+        }
+    }
+
+    #[cfg(feature = "trailing_ones")]
+    /// Returns the number of least-significant bits that are ones.
+    pub fn trailing_ones(&self) -> u64 {
+        if let Some(i) = self.data.iter().position(|&digit| !digit != 0) {
+            let ones: u64 = self.data[i].trailing_ones().into();
```
cuviper, line 2709: The `std` implementation of `trailing_ones` for primitives is literally just `(!self).trailing_zeros()`, so I think you can do the same here to avoid the feature / question of rustc support.
```suggestion
            let ones: u64 = (!self.data[i]).trailing_zeros().into();
```

src/biguint.rs
```diff
@@ -2689,6 +2689,34 @@ impl BigUint {
...
+            i as u64 * u64::from(big_digit::BITS) + ones
+        } else {
+            0
```
cuviper, line 2712: This misses the edge case where all digits are full ones (`uN::MAX`), so none match `!digit != 0`.
```suggestion
            self.data.len() as u64 * u64::from(big_digit::BITS)
```

cuviper (changes requested):

BartMassey:
Thanks very much for the careful review and catching all the paper bag bugs. I think this gets all your suggested changes…

cuviper:
Thanks!

bors r+

BartMassey:
Thank you!

PR #176, opened 2020-11-14
silenced a bunch of misleading Clippy warnings

BartMassey:
* There were a bunch of code quality Clippy warnings that didn't apply for various reasons.

* There was an ok_or() that could have been an ok_or_else() that Clippy caught. Turns out it's for a ZST, so doesn't matter, but was "fixed" anyway to silence Clippy.

This will hopefully make "cargo clippy" more useful for the next person.

### DIFF
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -349,6 +349,7 @@ fn bigint_from_slice(slice: &[BigDigit]) -> BigInt {
 
 /// Three argument multiply accumulate:
 /// acc += b * c
+#[allow(clippy::many_single_char_names)]
 fn mac3(acc: &mut [BigDigit], b: &[BigDigit], c: &[BigDigit]) {
     let (x, y) = if b.len() < c.len() { (b, c) } else { (c, b) };
 
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -2707,7 +2707,9 @@ impl TryFrom<&BigInt> for BigUint {
 
     #[inline]
     fn try_from(value: &BigInt) -> Result<BigUint, TryFromBigIntError<()>> {
-        value.to_biguint().ok_or(TryFromBigIntError::new(()))
+        value
+            .to_biguint()
+            .ok_or_else(|| TryFromBigIntError::new(()))
     }
 }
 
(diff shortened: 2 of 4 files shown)


BartMassey:
@cuviper These diffs are a bit gratuitous but I think they'll be a net positive. I chose to silence Clippy rather than change working code any more than necessary. Hopefully cutting down the Clippy noise will help the next person.

src/bigint.rs
```diff
@@ -2707,7 +2707,9 @@ impl TryFrom<&BigInt> for BigUint {
 
     #[inline]
     fn try_from(value: &BigInt) -> Result<BigUint, TryFromBigIntError<()>> {
-        value.to_biguint().ok_or(TryFromBigIntError::new(()))
+        value
+            .to_biguint()
+            .ok_or_else(|| TryFromBigIntError::new(()))
```
cuviper, line 2712: I think clippy is being overly fussy here since `TryFromBigIntError<()>` is a ZST, and `new` isn't doing anything significant, but I guess there's no harm in wrapping this. Another option would be to write this as a direct struct initialization for the ZST, but I think it won't matter for performance either way.

src/biguint.rs
```diff
@@ -1811,10 +1811,15 @@ impl Roots for BigUint {
...
 }
 
+#[allow(clippy::useless_conversion)]
 fn high_bits_to_u64(v: &BigUint) -> u64 {
     match v.data.len() {
         0 => 0,
-        1 => u64::from(v.data[0]),
+        1 =>
+        {
+            #[allow(clippy::useless_conversion)]
```
cuviper, line 1820: Why is the attribute repeated here?

cuviper (commented):
:+1: for making clippy more useful here.

src/algorithms.rs
```diff
@@ -347,6 +347,7 @@ fn bigint_from_slice(slice: &[BigDigit]) -> BigInt {
     BigInt::from(biguint_from_vec(slice.to_vec()))
 }
 
+#[allow(clippy::many_single_char_names)]
```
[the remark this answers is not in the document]
BartMassey: Got all these, I think.

src/bigint.rs
BartMassey: Ah. Yeah, I think it's probably the cleanest way to shut Clippy up.

src/biguint.rs
BartMassey: Mistake. Fixed.

PR #178, opened 2020-11-20
Normalize after all shrinking operations.

tczajka:
This guarantees that memory use will always stay O(n) for n-digit numbers.

### DIFF
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -520,6 +520,7 @@ impl Zero for BigUint {
     #[inline]
     fn set_zero(&mut self) {
         self.data.clear();
+        self.normalize();
     }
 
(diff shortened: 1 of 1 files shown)


src/biguint.rs
```diff
@@ -520,6 +520,7 @@ impl Zero for BigUint {
     #[inline]
     fn set_zero(&mut self) {
         self.data.clear();
+        self.normalize();
```
cuviper, line 523: Hmm, the shrink condition is `len < capacity / 4`, which means `set_zero` would only keep its allocation if it started with 3 or less. That's not much if someone is trying to avoid reallocation, but I'm also not sure anyone really cares...

PR #179, opened 2020-12-01
Create feature to override with u64, which would force u64 emulation

jon-chuang:
LLVM should handle the u64 emulation on 32-bit platforms just fine, if this is the desire of the lib user.


cuviper:
I consider the digit size to be an implementation detail. Using `u64` should always _work_, as far as codegen goes, but it's expected to be slower on 32-bit targets. Can you explain why a user would want to control this?

cuviper:
I'm closing this for lack of response.

PR #182, opened 2020-12-22
Add invmod algorithm

unseddd:
Add invmod algorithm using Extended Euclidean Algorithm

Based on implementation from [wolfSSL](https://github.com/wolfSSL/wolfssl/blob/master/wolfcrypt/src/tfm.c#L1038) (`fp_invmod` directly below `fp_invmod_slow`)

### DIFF
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -900,6 +900,10 @@ fn biguint_shr2(n: Cow<'_, BigUint>, digits: usize, shift: u8) -> BigUint {
     biguint_from_vec(data)
 }
 
+pub(crate) fn invmod(base: &BigInt, modulus: &BigInt) -> BigInt {
+    base.invmod(modulus)
+}
+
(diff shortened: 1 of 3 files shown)


cuviper:
I'm sorry, but I can't accept code that was translated from GPL-licensed code:
https://www.gnu.org/licenses/gpl-faq.html#TranslateCode

There's a generic implementation in rust-num/num-integer#10, and assuming we move that to a trait (on `Integer` or something new) then we can optimize it further here for `num-bigint`. (But that still must be clean from incompatibly-licensed code!)

unseddd:
> I'm sorry, but I can't accept code that was translated from GPL-licensed code

No worries, I wasn't aware this library is not compatible with GPL. I was using similar code in another project, so figured I would try porting it upstream.

Just trying to help out.

> There's a generic implementation in rust-num/num-integer#10

Cool! Hope it gets implemented for Big(U)Int, really helpful for crypto stuff.

PR #183, opened 2021-01-01
Get/set n-th bit of `BigUint` and `BigInt`

### LINKED ISSUE #172: How could I get/set n-th bit of `BigInt`?
Would you do me a favor? I need some help.

I'd like to get and set $n$-th bit of `BigInt`. How could I do it in a fast way?

For example, [ramp has such methods; `bit` and `set_bit`](https://docs.rs/ramp/0.5.9/ramp/int/struct.Int.html#method.bit).

I know that I can do it by bitwise operation such as `(x >> bit) & BigInt::one() == BigInt::one()` or so on. But I need faster way.

I would be grateful if you could give me advice.

janmarthedal:
This PR implements `bit` and `set_bit` for `BigUint` and `BigInt`.

The method names have been chosen to match those of Ramp (https://docs.rs/ramp/0.5.9/ramp/int/struct.Int.html#method.bit).

For `BigInt` the implementation uses the number's two's complement representation when the number is negative. This matches what libraries like Ramp or languages like Python do.

Resolves #172

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -3258,15 +3258,22 @@ impl BigInt {
     /// Returns whether the bit in position `bit` is set,
     /// using the two's complement for negative numbers
     pub fn bit(&self, bit: u64) -> bool {
-        // Let the binary representation of a number be
-        //   ... 0  x 1 0 ... 0
-        // Then the two's complement is
-        //   ... 1 !x 1 0 ... 0
-        // where !x is obtained from x by flipping each bit
-        if bit >= u64::from(big_digit::BITS) * self.len() as u64 {
-            self.is_negative()
-        } else if self.is_negative() && bit > self.data.trailing_zeros().unwrap() {
-            !self.data.bit(bit)
+        if self.is_negative() {
+            // Let the binary representation of a number be
+            //   ... 0  x 1 0 ... 0
+            // Then the two's complement is
+            //   ... 1 !x 1 0 ... 0
+            // where !x is obtained from x by flipping each bit
+            if bit >= u64::from(big_digit::BITS) * self.len() as u64 {
+                true
+            } else {
+                let trailing_zeros = self.data.trailing_zeros().unwrap();
+                match bit.cmp(&trailing_zeros) {
+                    Ordering::Less => false,
+                    Ordering::Equal => true,
+                    Ordering::Greater => !self.data.bit(bit),
+                }
+            }
         } else {
             self.data.bit(bit)
         }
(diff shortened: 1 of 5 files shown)


src/bigint.rs
```diff
@@ -3254,6 +3254,70 @@ impl BigInt {
...
+                // computing trailing_zeros can be avoided when the bit to set/clear is outside
+                // the represented digits
+                if bit >= bits_per_digit * self.len() as u64
+                    || bit > self.data.trailing_zeros().unwrap()
+                {
+                    self.data.set_bit(bit, !value);
+                } else {
+                    // This is the general case that basically corresponds to what `bitor_neg_pos`
+                    // (when setting bit) or `bitand_neg_pos` (when clearing bit) does, except there
+                    // is no need to explicitly iterate over the digits of the right-hand side
```
cuviper, line 3299: I have to admit, that bit-twiddling has fallen out of my memory by now. Your adaptation looks appropriate, but I can't help feeling that there must be a way to shorten the scan both before and after the target bit. That is, before only needs to figure out the carry state, and after might be able to stop on zero carry. I'm not sure.

Maybe this is a good candidate for some fuzzing? There are tests of this sort in `ci/big_quickcheck`.

cuviper (commented):
Thanks! The two's complement stuff is a bit hairy, as expected, but I don't see any problem with it.

src/bigint.rs
janmarthedal: You are absolutely correct, in most cases it will iterate over far more than necessary. When writing the code, I was in doubt of whether I should go for more efficient and complicated code, or sub-optimal code that was fairly straightforward and shorter. I went for the latter, but I sort of regret that now. I am currently rewriting the `set_bit` method, so stay tuned!

janmarthedal:
@cuviper I have now updated the PR with a more efficient implementation of `BigInt::set_bit`. Basically, I have made sure that only the relevant digits will be iterated over instead of all of them, as you also pointed out. In a special benchmark it saved about 70% in runtime.

I have also implemented a special case for `bit < trailing_zeros && value` which also saves some runtime compared to the case `bit == trailing_zeros && !value`, but only about 5-10%. The two cases can be merged together to make the `set_bit` a bit shorter.

mkaczanowski:
@cuviper it's been a while since the last review :) Is this PR ready? (asking, since I am waiting for this feature)

cuviper:
I added a couple tweaks to the edge cases of `u64` to `usize`, but otherwise it looks good!

bors r+

PR #185, opened 2021-01-05
Bump rand to 0.8

Gelbpunkt:
This updates rand to 0.8, which is required when working with a 0.8 version in the package since the version needs to match for the trait.


cuviper:
Hi, and thanks! This will need updates in `ci/big_rand` as well.

Updating a public dependency is a breaking change, so I would like to wait for `quickcheck` and `arbitrary` to update too, and we can bump them all in one release. It looks like both of those are getting ready for their 1.0 releases.
(The update PRs can be handled separately though.)

cuviper:
Thanks for your patience! I'm getting ready for 0.4 now...

bors r+

PR #188, opened 2021-01-21
Add optimized extended binary gcd algorithm

unseddd:
Since `BigInt` uses potentially very large numbers, implement the optimized extended binary gcd algorithm from the Handbook of Applied Cryptography.

Blocked on [num-integer/#37](https://github.com/rust-num/num-integer/pull/37)


unseddd:
Close in favor of [num-integer/#38](https://github.com/rust-num/num-integer/pull/38)

PR #192, opened 2021-02-24
Add iter_u32_digits , iter_u64_digits, to_u64_digits to both BigInt and BigUint

cuviper:
This is a rebase of #158, and I also addressed my last review comments there. I did rename the iterators to `U32Digits` and `U64Digits`, but I left the `iter_` prefix on the methods so they're not confused with the similar `to_` methods.


cuviper:
cc @Speedy37

PR #202, opened 2021-04-11
Add `truncate_*` operations on `BigInt` and `BigUint`

digama0:
Implements https://github.com/rust-num/num/issues/404

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -29,6 +29,7 @@ mod bits;
 mod convert;
 mod power;
 mod shift;
+mod truncate;
 
 #[cfg(any(feature = "quickcheck", feature = "arbitrary"))]
 mod arbitrary;
--- a/src/bigint/truncate.rs
+++ b/src/bigint/truncate.rs
@@ -1,0 +1,31 @@
+use super::{BigInt, Sign};
+
+macro_rules! impl_truncate_bigint {
+    ($T:ty, $truncate_N:ident) => {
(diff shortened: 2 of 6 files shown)


digama0:
Hi @cuviper, gentle reminder that this PR exists and could use review.

cuviper:
I wonder if there's a way we can do this generically, rather than so many new methods, something like:

```rust
impl BigUint {
    pub fn truncate<T>(&self) -> T
    where
        T: PrimInt,
        u128: AsPrimitive<T>,
    {
        todo!("truncate as u128").as_()
    }
}

impl BigInt {
    pub fn truncate<T>(&self) -> T
    where
        T: PrimInt,
        i128: AsPrimitive<T>,
    {
        todo!("truncate as i128").as_()
    }
}
```

The `u128`/`i128` distinction here doesn't really matter, but it looks appropriate.

Or maybe `T: PrimInt` is sufficient to build `T` directly? Not sure -- I think signs may be tricky.

digama0:
Check out the latest commit for a version that puts all the functions under a trait. It's not `PrimInt`, but a `TruncateFrom` trait; the actual front-end functions are the inherent `truncate` methods on `BigInt` and `BigUint` (which allows for turbofishing and also avoids the need to import the trait at use sites). `BigInt::truncate` has the cumbersome bound `TruncateFrom<BigUint> + Zero + WrappingNeg` but users shouldn't have to deal with it directly.

PR #208, opened 2021-07-22
Add `DoubleEndedIterator` to `U32Digits` and `U64Digits`

PatrickNorton:
Since `U32Digits` and `U64Digits` do not implement `DoubleEndedIterator`, there is currently no easy way to iterate over the digits in a big-ended fashion.


cuviper:
> Since `U32Digits` and `U64Digits` do not implement `DoubleEndedIterator`,

Did you explore what it would take to add that implementation instead?

PatrickNorton:
I'd thought about it a little but went with this instead -- doing a quick mockup, it looks like that won't be as difficult as I first thought, so I'll give it a go.

PatrickNorton:
It's definitely a lot cleaner than I expected; I think we should go that route instead.

PatrickNorton:
@cuviper Thoughts on the new implementation?

cuviper:
It looks good to me, thanks!

bors r+

cuviper:
Hmm, that failure with `arbitrary` is not your fault -- I'll see what I can do about this...

Issue #36, opened 2018-03-19
Base on SmallVec instead of Vec


kvark:
It would be great to allow for `BigInt` and alike to be used in constant expressions. This is not possible with `Vec`, but if we use `SmallVec` instead, we might be able to make this possible. See https://github.com/servo/rust-smallvec/issues/86

cuviper:
AFAICT, `smallvec` only tests with the latest stable rustc, which is not a great fit for num-bigint since we hold a stronger guarantees about rustc requirements -- currently targeting 1.15 for the development on num-bigint 0.2.

It might be feasible to implement our own internal `SmallVec` though, since we'd only need a smaller subset of the normal `Vec`-like API.  It will need to be thoroughly benchmarked -- avoiding the heap is a benefit, but the added conditional branches could hurt.  It's not an obvious win.

On constants -- I doubt we'll change the traits as you're proposing in rust-num/num-traits#54, but at least with a small-vec implementation, it should be possible to add such constants on the `BigUint`/`BigInt` types themselves.

kvark:
> smallvec only tests with the latest stable rustc, which is not a great fit for num-bigint since we hold a stronger guarantees about rustc requirements -- currently targeting 1.15 for the development on num-bigint 0.2.

True, but we can easily fix that. `SmallVec` belongs to Servo org, which I'm a member of. I'm sure we can figure this out ;)

> It might be feasible to implement our own internal SmallVec though, since we'd only need a smaller subset of the normal Vec-like API. It will need to be thoroughly benchmarked -- avoiding the heap is a benefit, but the added conditional branches could hurt. It's not an obvious win.

I see. Alternative approach would be to have some sort of a generic parameter for allocation of `BigInt`, making sure that instances created from different allocations can still inter-operate. It would basically mean lifting the mentioned conditional branch into the type system and transitioning the cost from run-time to compile time (plus, a bit of the API complexity).

clanehin:
I noticed that this is marked "help wanted." I might be interested in taken it on. If so I would imagine implementing a SmallVec version of one of the BigInts first, as a relatively easy exploratory option (and benchmarking) and then taking some course of action based on what we learn from that and whether or not SmallVec will adapt to meet num-bigints testing requirements, etc.

If that sounds reasonable, I'll get started. I have a few hours free each week so it might take several weekly iterations.

clanehin:
[...]

PR #210, opened 2021-07-28
Use SmallVec to optimize for small integer sizes.

lemmih:
This is a proof of concept. I'd like to discuss possible designs and pathways to getting something like this merged.

# Motivation

Using `BigInt` with mostly small (<= word-sized) integers is prohibitively expensive. One numerically intense algorithm from rgeometry is 1000 times slower when using `BigRational` instead of a fixed-precision number.

# Performance

`num-bigint` performs quite well for large input and is competitive with the GMP. Detailed benchmarks can be found here: https://github.com/Lemmih/num-criterion

So why do small integers perform so poorly? Mostly because allocating new memory is significantly more expensive than doing a single addition or multiplication. There are also a few cases where we can use specialized algorithms. For example, there's a particularly fast `gcd` algorithm that only works on word-sized integers.

# SmallVec with two inlined BigDigits

`BigUint` is a `Vec<BigDigit>`. As such, it has a size of 3 words in addition to the actual digits. Switching to a `SmallVec<[BigDigit; 2]>` will not increase the size of a `BigUint` but allows for two BigDigits to be inlined and thus not require a heap allocation.

# Summary of benchmarks:

I've run benchmarks on an M1 Apple MacBook Air and a 3950X AMD desktop. The results I get from these platforms are quite different and I'd love it if other people could help out by running the benchmark on their machines. For instructions, see: https://github.com/Lemmih/num-criterion

 - **GCD**: Slightly faster across all bit sizes. This is mainly because I wrote a more efficient version of `shr_assign`. The previous version interacted poorly with `SmallVec`.
 - **Mul**: Significantly faster for bit sizes <=64. Slightly slower for 128-bit integers. No significant difference for larger integers.
 - **Div**: Significantly faster for bit sizes <=128. No significant difference for larger integers.
 - **Add**: Significantly faster for bit sizes <=128. No significant difference for larger integers.
 - **Clone**: Significantly faster for bit sizes <=128. No significant difference for larger integers.

# Other optimizations

## Specialize for integers that fit in 128 bits.

Ideally, addition, multiplication, and division for integers that fit in 128 bits would be as fast as cloning. Copying the bytes is significantly slower than doing the arithmetic.

 * Signed integer addition/subtraction is quite naive. It is comparing the absolute numbers before subtracting/adding. This has to be improved before specialization will have any noticeable effect. Surely we can get within 2x of GMP.

## Use better GCD implementation.

The GCD implementation in `num-integer` is fairly slow and can be significantly improved by changing a line or two. On my system, this leads to a 10x speed improvement for BigInts when the integer fits in 64 bits.

# Other oddities

* Rug and Ramp are really good at cloning data. How can they clone large integers at more than twice the speed of `num-bigint`? Aren't we all just calling 'memcpy'?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,6 @@ edition = "2018"
 
 [features]
 default = ["std"]
-union = ["smallvec/union"]
 std = ["num-integer/std", "num-traits/std"]
 
 [package.metadata.docs.rs]
@@ -39,7 +38,7 @@ harness = false
 name = "shootout-pidigits"
 
 [dependencies]
-smallvec = { version = "1.6.1" }
+smallvec = { version = "1.6.1", optional = true, features = ["union"] }
 
 [dependencies.num-integer]
 version = "0.1.42"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ edition = "2018"
 
 [features]
 default = ["std"]
+union = ["smallvec/union"]
 std = ["num-integer/std", "num-traits/std"]
 
 [package.metadata.docs.rs]
@@ -38,6 +39,7 @@ harness = false
 name = "shootout-pidigits"
 
 [dependencies]
+smallvec = { version = "1.6.1" }
 
 [dependencies.num-integer]
 version = "0.1.42"
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -14,6 +14,8 @@ use core::{i64, u64};
 use num_integer::{Integer, Roots};
 use num_traits::{Num, One, Pow, Signed, Zero};
 
+use smallvec::SmallVec;
+
 use self::Sign::{Minus, NoSign, Plus};
 
 use crate::big_digit::BigDigit;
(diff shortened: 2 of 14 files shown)


Lemmih:
This PR is the same as the previous one with the exception of a few optimisations. The optimisations weren't necessary; they just proved that the `SmallVec` backend could be as efficient as `Vec`.

Lemmih:
Oh, and I added back the 'shr' optimisation because the regression was particularly egregious.

Lemmih:
@cuviper Thoughts?

cuviper:
How much of a difference does the union make? I prefer that conceptually because we don't increase the size that way, but in past experiments ISTR it had mixed impact on performance overall.

If the union performance is good, then I think I'd prefer having a feature that toggles between `Vec` or `SmallVec` (w/ union), where the latter requires MSRV 1.49, but we'd keep the same struct size either way. I think the global type alias would make toggling easy, as long as we're not using any API that happens to be different between them. What do you think?

Lemmih:
The performance difference between union and no-union is minor and dwarfed by the performance boost from avoiding heap allocations.

I'll update the PR to include a feature switch for `SmallVec`.

Rudxain:
Should this PR be marked as "Closes #36"?

x10an14:
What's stalled this PR? Atm there are merge conflicts, but I assume it's reasonable to assume that's not why it stalled originally.

cuviper:
I went with a bespoke enum in #307.

Issue #184, opened 2021-01-03
impl Not for BigUint


Lichtso:
https://docs.rs/num-bigint/0.3.1/num_bigint/struct.BigInt.html?search=not

Seems like there is only an implementation of `not` for `BigInt`.

cuviper:
Conceptually, `BigUint` has infinite leading zeros. There's no clear answer how far we should flip bits for `Not`. Even if we just flipped the current number of inner `BigDigit` (which is arch-dependent), this may be inconsistent in the round trip, `!!x`, when we uphold the invariant that leading zero digits are trimmed. For that matter, `BigUint::zero()` has no digits at all, so would its `not()` also be zero?

With `BigInt`, we conceptually have infinite leading bits either way, zeros for positive and ones for negative, like 2's complement. So we have a way to implement `Not` that is consistent with signed primitive integers. That is, 2's complement `-x = !x + 1`, so `!x = -x - 1`.

Lichtso:
In theory one could define the MSB to be implicitly extended towards infinity (like sign extension).
So `0 ... 0001`
would be inverted to `1 ... 1110`
and back to `0 ... 0001`.

But that probably breaks a lot of the other assumptions and operations.

cuviper:
Do you have a use for `Not` on a `BigUint`? Perhaps we can solve that another way. Maybe we should also add my explanation or similar to the docs somewhere.

PR #211, opened 2021-08-08
Add bitwise operations with primitive ints

### LINKED ISSUE #205: Add bitwise ops with primitive ints
Things like `impl BitAndAssign<u8> for BigInt` are missing.

PatrickNorton:
Fixes #205.

This does not add bitwise-or and bitwise-xor operations between signed integers and BigUint, because there is no good way to represent negative numbers with a BigUint (see #184). Bitwise-and operations were added because the result of such an operation will always be positive.

### DIFF
--- a/src/bigint/bits.rs
+++ b/src/bigint/bits.rs
@@ -1,13 +1,14 @@
 use super::BigInt;
-use super::Sign::{Minus, NoSign, Plus};
+use super::Sign::{self, Minus, NoSign, Plus};
 
+use crate::{IsizePromotion, UsizePromotion};
 use crate::big_digit::{self, BigDigit, DoubleBigDigit};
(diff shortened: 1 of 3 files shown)


cuviper:
> Bitwise-and operations were added because the result of such an operation will always be positive.

Just because we can, doesn't mean we should, and this would be the only operator that mixed `BigUint` with signed primitives.
(Well, except for shifts, but the standard library already set the precedence for that oddness...)

PatrickNorton:
You're right; I hadn't realized that the other operations would have such an issue until after I'd written the `BitAnd` implementation, and I figured it'd be better to ask before deleting it. I'll happily do so now if it doesn't make sense to include.

cuviper:
I still think the mixed signed-ness is too confusing, let's drop that.

PR #212, opened 2021-08-08
Add `BigUint::{is, next}_power_of_two()`

PatrickNorton:
These are two more integer methods we don't have on `BigUint`, which would be nice to have. 

Unresolved questions: 
- [x] Currently, `next_power_of_two` takes `self` by value instead of reference, to prevent unnecessary clones. Is that the right decision?
- [x] If we decide to have `next_power_of_two` take `self` by reference, should we add a `to_next_power_of_two` method that takes it by value?

### DIFF
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -954,6 +954,40 @@ impl BigUint {
             self.normalize();
         }
     }
+
+    /// Returns `true` if and only if `self == 2^k` for some `k`.
+    pub fn is_power_of_two(&self) -> bool {
+        self.count_ones() == 1
+    }
+
+    /// Returns the smallest power of two greater than or equal to `self`.
+    pub fn next_power_of_two(mut self) -> BigUint {
+        match self.data.len() {
+            0 => BigUint::one(),
+            1 => {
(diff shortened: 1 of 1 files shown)


cuviper:
On naming, there are these guidelines between `as_`, `to_`, and `into_`:
https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

There is also the precedent of the `next_power_of_two` method on the primitive integers, which is not prefixed at all and takes `self` by value. But those integers are trivially `Copy`, so they almost never bother with borrowed `&self`.

There are also a lot of `num-traits` methods that take `&self` for the sake of types like `BigUint`, even when the primitive types would take `self` by value in almost every case.

So if we want to support both owned and borrowed, I could see either of these options working well:
- `next_power_of_two(self)` and `to_next_power_of_self(&self)`
- `next_power_of_two(&self)` and `into_next_power_of_self(self)`

PatrickNorton:
I'd think it's better to go with the by-value as the default; it seems better to promote the more efficient method as the default and let people switch out of it when they need it.

PR #214, opened 2021-08-12
Add exponential formatting to BigInt

### LINKED ISSUE #6: implement LowerExp and UpperExp
These could be helpful for printing large numbers approximately, especially if we can figure out how to do it quickly, without full digit-by-digit division.

cc rust-num/num#259

PatrickNorton:
Should close #6.

### DIFF
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -33,6 +33,7 @@ mod serde;
 
 pub(crate) use self::convert::to_str_radix_reversed;
 pub use self::iter::{U32Digits, U64Digits};
+use crate::Sign;
 
 /// A big unsigned integer type.
 pub struct BigUint {
@@ -156,10 +157,10 @@ impl fmt::LowerExp for BigUint {
                 if as_f64.is_finite() {
                     fmt::LowerExp::fmt(&as_f64, f)
                 } else {
-                    fmt_exp_slow(self, f, false)
+                    fmt_exp_slow(self, f, false, true)
                 }
             } else {
-                fmt_exp_slow(self, f, false)
+                fmt_exp_slow(self, f, false, true)
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/biguint.rs
```diff
@@ -143,6 +144,72 @@ impl fmt::Octal for BigUint {
...
+}
+
+impl fmt::UpperExp for BigUint {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Option::Some(small) = self.to_u64() {
+            fmt::UpperExp::fmt(&small, f)
+        } else {
+            // The max digits that can fit into a f64, which ensures
+            // that printing won't have rounding errors.
+            const MAX_FLOAT_DIGITS: usize = 14;
```
cuviper, line 176: `f64::DIGITS` is 15, though it does say "approximate". Perhaps we should us that, minus one?

src/biguint.rs
```diff
@@ -143,6 +144,72 @@ impl fmt::Octal for BigUint {
...
+    let precision = f.precision().unwrap_or(16);
+    let digits = to_str_radix_reversed(x, 10);
+    let mut buffer = String::with_capacity(precision + 5);
+    debug_assert!(digits.len() > 1, "Values with fewer digits should use direct formatting");
+    let mut iter = digits.iter().rev();
+    buffer.push(*iter.next().unwrap() as char);
+    buffer.push('.');
+    for &ch in iter.take(precision) {
+        buffer.push(ch as char);
+    }
```
cuviper, line 201: Don't we need rounding if there were more digits than the precision required?

src/biguint.rs
```diff
@@ -143,6 +144,72 @@ impl fmt::Octal for BigUint {
...
+    // Add trailing zeroes when explicit precision given
+    if f.precision().is_some() {
+        for _ in 0..precision.saturating_sub(digits.len() - 1) {
+            buffer.push('0');
+        }
+    }
+    buffer.push(if upper { 'E' } else { 'e' });
+    write!(buffer, "{}", digits.len() - 1)?;
+    f.pad_integral(is_nonneg, "", &buffer)
```
cuviper, line 210: `pad_integral` feels weird in this case, but I'm not sure it's wrong. The standard library uses its internal `pad_formatted_parts` for exponential integers, but at a glance I don't see what would be different.

src/biguint.rs
PatrickNorton: I subtracted 1 to ensure the rounding wouldn't turn out weird, but forgot to update the docs. I didn't know `f64::DIGITS` existed, so I'll update to take that into account.

src/biguint.rs
PatrickNorton: It's definitely weird, but it does seem to do the right thing. I've added a comment to that effect.

src/biguint.rs
PatrickNorton: Good point! Added.

cmpute:
Would this PR be merged soon? I'm looking for similar functionality on this.

PR #225, opened 2021-10-19
Refactor conversion methods to avoid floating points on `no_std`

archseer:
This might be better suited behind a new feature gate? We're trying to use `num-bigint` in a WASM project that's running on [CosmWasm](https://docs.cosmwasm.com/docs/0.16/). The code compiles fine, but when trying to execute the program it's rejected by the VM because it refuses to run bytecode containing floating point operations.

Turns out the codebase already does a good job of avoiding such operations with the `std` feature disabled, but it's still being used in `from_radix_digits_be`/`_le` to estimate the final result size. I've switched this out on the no std code path and it seems to run fine with this change.


cuviper:
Hi, and sorry for the review silence. This looks totally reasonable to avoid floats on `no_std`, especially since it's just an estimate anyway. Thanks!

bors r+

PR #229, opened 2021-11-07
Added `BigUInt::factorial`

PatrickNorton:


src/biguint/factorial.rs
```diff
@@ -0,0 +1,91 @@
...
+    1856156927625,      // 19! = 1856156927625 << 16
+    9280784638125,      // 20! = 9280784638125 << 18
+    194896477400625,    // 21! = 194896477400625 << 18
+    2143861251406875,   // 22! = 2143861251406875 << 19
+    49308808782358125,  // 23! = 49308808782358125 << 19
+    147926426347074375, // 24! = 147926426347074375 << 22
+];
+
+pub fn factorial(x: usize) -> BigUint {
+    if let Option::Some(&x) = SMALL_FACTORIALS.get(x) {
```
pierwill, line 60:
```suggestion
    if let Some(&x) = SMALL_FACTORIALS.get(x) {
```

src/biguint.rs
```diff
@@ -954,6 +955,15 @@ impl BigUint {
             self.normalize();
         }
     }
+
+    /// Computes the factorial of the `BigUInt`.
+    pub fn factorial(&self) -> BigUint {
+        // usize::MAX! > 2**usize::MAX, so just error w/o running the system out of memory
+        match self.to_usize() {
```
Rudxain, line 962: The growth of `x!` is similar to `x**x`, so the input should be constrained to less than half the bit-width (`usize::MAX.isqrt())` or `1 << (usize::BITS / 2))`)

Rudxain (changes requested):

PR #234, opened 2022-01-14
[docs] Use assignment instead of replace

icecream17:
The line moves f2 into f1 using `replace` but that can be done with assignment

(I'm a beginner to rust so it's very likely I've missed something)


cuviper:
Hi, and sorry for the review silence. You're right, this is a good cleanup in the example!

bors r+

PR #236, opened 2022-02-23
Use inline asm! for x86 DIV

cuviper:


cuviper:
bors try

cuviper:
While this does improve benchmarks here and in my own programs, one downside is that this `asm!` is opaque to LLVM, so it doesn't get a chance to optimize division by a constant (scalar) value into "magic" multiplication.

cuviper:
On further reflection, the *measured* gains should outweigh theoretical optimization losses.

Maybe someday we could try [`is_val_statically_known`](https://doc.rust-lang.org/std/intrinsics/fn.is_val_statically_known.html) to detect divisors that might be optimizable.

PR #237, opened 2022-03-10
Removed unnecessary lifetime annotations

OliveIsAWord:
Out of 110 lifetime generics in 20 files, only 23 in 4 files were necessary for compilation. This should make code more readable and not affect behavior in any way.

Hopefully this works. I tested it locally on both 1.31 and the latest version 1.59.

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -196,7 +196,7 @@ impl Not for BigInt {
     }
 }
 
-impl<'a> Not for &'a BigInt {
+impl Not for &BigInt {
     type Output = BigInt;
 
     fn not(self) -> BigInt {
@@ -342,7 +342,7 @@ impl Neg for BigInt {
     }
 }
 
-impl<'a> Neg for &'a BigInt {
+impl Neg for &BigInt {
     type Output = BigInt;
 
     #[inline]
--- a/src/bigint/addition.rs
+++ b/src/bigint/addition.rs
@@ -30,7 +30,7 @@ macro_rules! bigint_add {
     };
 }
 
-impl<'a, 'b> Add<&'b BigInt> for &'a BigInt {
+impl Add<&BigInt> for &BigInt {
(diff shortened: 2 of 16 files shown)


src/biguint/iter.rs
```diff
@@ -117,9 +117,9 @@ impl ExactSizeIterator for U32Digits<'_> {
 }
 
 #[cfg(not(u64_digit))]
-impl<'a> U32Digits<'a> {
+impl U32Digits<'_> {
     #[inline]
-    pub(super) fn new(data: &'a [u32]) -> Self {
+    pub(super) fn new(data: &[u32]) -> Self {
```
cuviper, line 122: This one doesn't work on 32-bit targets, I guess for the same reason that you couldn't make this change on `cfg(u64_digit)`.

src/biguint/iter.rs
```diff
@@ -181,9 +181,9 @@ pub struct U64Digits<'a> {
 }
 
 #[cfg(not(u64_digit))]
-impl<'a> U64Digits<'a> {
+impl U64Digits<'_> {
     #[inline]
-    pub(super) fn new(data: &'a [u32]) -> Self {
+    pub(super) fn new(data: &[u32]) -> Self {
```
cuviper, line 186: This also fails on 32-bit, but it passes with `-> U64Digits<'_>` to have a different anonymous lifetime than the `impl`'s `Self`. But that doesn't work with the other case, for some reason...

cuviper (commented):
Thanks, this looks good, but for a gap in 32-bit coverage.

(edit: To be clear, 32-bit will run in full CI, after approval, but I also noticed that locally first.)

OliveIsAWord:
Right, silly oversight on my part. It should work now I believe.

cuviper:
Hi again -- I don't know why I didn't merge this before, but let's try now...

bors r+

Issue #146, opened 2020-04-16
Implement Euclidean division and modulo


betaveros:
I see we have floor division and the corresponding modulo operations via the Integer trait. However methods for the subtly different Euclidean division and modulo operations were added to the core Rust primitives in [RFC 2169](https://github.com/rust-lang/rfcs/blob/master/text/2169-euclidean-modulo.md) and it might be good to have them on BigInt for consistency. (If I have missed the implementation, please let me know. I am fairly new to Rust.)

cuviper:
I'm open to it. It would be nice to sort out a common trait for this first (rust-num/num-traits#159), then `num-bigint` can implement that.

PR #245, opened 2022-07-16
Implement Euclid division and remainder

jaybosamiya:
Now that a common trait for this has been sorted out (https://github.com/rust-num/num-traits/issues/159 implemented and merged in https://github.com/rust-num/num-traits/pull/195), we can now close https://github.com/rust-num/num-bigint/issues/146 by implementing the trait for `BigInt`s. This commit does just that.

### DIFF
--- a/src/bigint/division.rs
+++ b/src/bigint/division.rs
@@ -6,7 +6,7 @@ use crate::{IsizePromotion, UsizePromotion};
 
 use core::ops::{Div, DivAssign, Rem, RemAssign};
 use num_integer::Integer;
-use num_traits::{CheckedDiv, ToPrimitive, Zero};
+use num_traits::{CheckedDiv, CheckedEuclid, Euclid, ToPrimitive, Zero};
 
 forward_all_binop_to_ref_ref!(impl Div for BigInt, div);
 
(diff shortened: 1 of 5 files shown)


src/bigint/division.rs
```diff
@@ -446,3 +446,51 @@ impl CheckedDiv for BigInt {
...
+        }
+        Some(self.rem_euclid(v))
+    }
+}
+
+impl Euclid for BigInt {
+    #[inline]
+    fn div_euclid(&self, v: &BigInt) -> BigInt {
+        let q = self / v;
+        if self % v < BigInt::zero() {
```
cuviper, line 472: This should use `div_rem` so we don't have to divide twice.

Also, `is_negative()` and `is_positive()` would be more direct than all the comparisons to `zero()`.

cuviper (changes requested):
Hi, and sorry for review silence.

For completeness, can you also add the trivial implementations for `BigUint`?

jaybosamiya:
Thanks for the review! I've added commits based on the above requested changes, and the branch is now ready for another review.

cuviper:
I rebased this and added a couple compat fixes. Otherwise, looks good, thanks!

bors r+

cuviper:
bors retry

jaybosamiya:
Thanks for rebasing and merging!

PR #247, opened 2022-07-21
Update alternatives list

AaronKutch:
`rust-gmp`, `ramp`, and `apint` are all unmaintained so I removed them. I added a column for fundamental features the crates have and added some new rows.


cuviper:
Thanks! I've rebased and applied further updates.

bors r+

PR #248, opened 2022-07-31
Carry only single word in multiplication

calebsander:
The carry from a multiply or a multiply-accumulate is a single word,
so update the functions to clarify that. Use a temporary double word for the
computation. And rename "acc" to "carry" to distinguish from the accumulator.

Also use multiply instead of multiply-accumulate for string conversions.

And remove an unnecessary mask in get_lo().


cuviper:
Have you run benchmarks for this change? It has been pretty sensitive in the past, whether this ultimately coaxes good behavior out of LLVM.

cuviper:
Closing for inactivity; please feel free to reopen when ready.

PR #257, opened 2022-10-02
Add support for conversions to ICU4X FixedDecimal

ajtribick:
Feature-gated support for fallible conversion to the `FixedDecimal` type used with ICU4X, enabling localized formatting of `BigInt` and `BigUint`. I've used the `icu` feature as the gate here as I thought this would be friendlier than using `fixed_decimal` directly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,9 +17,10 @@ edition = "2018"
 [features]
 default = ["std"]
 std = ["num-integer/std", "num-traits/std"]
+icu = ["fixed_decimal"]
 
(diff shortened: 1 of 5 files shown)


cuviper:
I think this would probably work better from `fixed_decimal`'s side. You can use the public `to_radix_be` or `_le` from here, and internally construct the `FixedDecimal` without any ascii digit conversions. It should also be much simpler there knowing that it doesn't have to worry about any fractional or exponential parts.

ajtribick:
Ok, no worries.

PR #263, opened 2022-12-19
Use `u64_digit` for 128-bit targets

icedrocket:
`CARGO_CFG_TARGET_*` variables are always set when cargo calls the build script. If these variables are not set, it should be considered a bug.

This PR makes sure these environment variables aren't missing, and cleans up the code a bit.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -5,29 +5,31 @@ use std::io::Write;
 use std::path::Path;
 
 fn main() {
-    let pointer_width = env::var("CARGO_CFG_TARGET_POINTER_WIDTH");
-    let u64_digit = pointer_width.as_ref().map(String::as_str) == Ok("64");
+    let ptr_width = &*env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap();
+
+    let u64_digit = ["64", "128"].contains(&ptr_width);
     if u64_digit {
         autocfg::emit("u64_digit");
     }
+
     let ac = autocfg::new();
     let std = if ac.probe_sysroot_crate("std") {
         "std"
     } else {
         "core"
     };
+
     if ac.probe_path(&format!("{}::convert::TryFrom", std)) {
         autocfg::emit("has_try_from");
     }
 
-    if let Ok(target_arch) = env::var("CARGO_CFG_TARGET_ARCH") {
-        if target_arch == "x86_64" || target_arch == "x86" {
-            let digit = if u64_digit { "u64" } else { "u32" };
+    let arch = &*env::var("CARGO_CFG_TARGET_ARCH").unwrap();
 
-            let addcarry = format!("{}::arch::{}::_addcarry_{}", std, target_arch, digit);
-            if ac.probe_path(&addcarry) {
-                autocfg::emit("use_addcarry");
-            }
+    if ["x86_64", "x86"].contains(&arch) {
+        let digit = if u64_digit { "u64" } else { "u32" };
+
+        if ac.probe_path(&format!("{}::arch::{}::_addcarry_{}", std, arch, digit)) {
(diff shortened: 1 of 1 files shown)


cuviper:
> If these variables are not set, it should be considered a bug.

I'm not sure why that's so. It seems fine to me to make a conservative choice in case they're missing, for whatever reason.

icedrocket:
Yes, there is no rule that it must be specified, but most crates call unwrap when loading Cargo-related environment variables. Since we're only supporting crates through the Cargo package manager, we're better off taking a conservative approach.

cuviper:
Note that we both said "conservative" with different intent...

What is the *benefit* to being stricter about this? We risk breaking the build, however unlikely that may be, but there's no point in doing that when the code will still work fine with a fallback.

icedrocket:
I checked when these environment variables were added to Cargo. These variables were added in [late 2016](https://github.com/rust-lang/cargo/commit/23587ee744ebd0f85714529f9c956b10803c4be8), after Rust 1.0 was released. This is before Rust 1.14 was released, so these environment variables are guaranteed to be set in Rust 1.31 and later.

As you mentioned, not being strict about missing environment variables doesn't currently cause any problems building crates. However, the reason I created this PR was because missing environment variables could be a bug in Cargo. AFAIK, there is no other program other than Cargo that uses a build script written in Rust.

build.rs
```diff
@@ -5,29 +5,31 @@ use std::io::Write;
 use std::path::Path;
 
 fn main() {
-    let pointer_width = env::var("CARGO_CFG_TARGET_POINTER_WIDTH");
-    let u64_digit = pointer_width.as_ref().map(String::as_str) == Ok("64");
+    let ptr_width = &*env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap();
+
+    let u64_digit = ["64", "128"].contains(&ptr_width);
```
cuviper, line 10: FWIW, I do agree that 128-bit targets should use `u64_digit` -- that's a positive change we can make. (even though there are no such targets yet, AFAIK)

icedrocket: I'll create a new PR to reflect what you mentioned.

icedrocket:
Now, thinking back to what I did a few weeks ago, I tried to keep my code as concise as possible. A syntax included in the latest Rust was not included in Rust 1.31, making it difficult to express in one line. For example:

```rust
let ptr_width = env::var("CARGO_CFG_TARGET_POINTER_WIDTH").ok();

let u64_digit = matches!(ptr_width.as_deref(), Some("64" | "128"));

// ...
```

```rust
let arch = env::var("CARGO_CFG_TARGET_ARCH").ok();

if let Ok(arch @ ("x86_64" | "x86")) = arch.as_deref() {
    //...
}
```

I couldn't use these patterns to continue supporting version 1.31, so I thought about how to express them concisely. One of the methods I've used to do this is just unwrap the Result. If we don't unwrap the Result, the expression becomes longer:

```rust
let u64_digit = ptr_width
    .as_ref()
    .map_or(false, |x| ["64", "128"].contains(&&**x));
```

or

```rust
let u64_digit = match ptr_width.as_ref().map(|x| &**x) {
    Some("64" | "128") => true,
    _ => false,
};
```

cuviper:
There's no need to golf this, especially since it shouldn't need much change. Either of those longer options look fine, although you'll have to split that pattern too, `Some("64") | Some("128")`.

icedrocket:
I am trying to add a new commit, but it is not reflected in the pull request. I contacted GitHub Support and am waiting for a response.

icedrocket:
I'll just create a new pull request.

icedrocket:
I'll just create a new pull request.

PR #271, opened 2023-04-27
Implement ties-to-even for Big[U]Int::to_{f32,f64}

dramforever:
This makes the rounding consistent with primitive types, which, per IEEE-754, uses nearest-ties-to-even rounding.

See: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a6ab9b18a4982f881cd594061f096f8c

```rust
#[test]
fn test() {
    use num::{BigInt, ToPrimitive};
    let a: i128 = (1 << 120) + (1 << (120 - 53));
    let b: i128 = 1 << (120 - 60);
    assert_ne!(a as f64, (a + 1) as f64);
    assert_eq!((a + b) as f64, (a + 1) as f64);

    assert_eq!(BigInt::from(a).to_f64().unwrap(), a as f64);
    
    // This works because the b is among the high 64 bits
    assert_eq!(BigInt::from(a + b).to_f64().unwrap(), (a + b) as f64);

    // This fails because 1 is just thrown away
    assert_eq!(BigInt::from(a + 1).to_f64().unwrap(), (a + 1) as f64);
}
```

### DIFF
--- a/tests/bigint.rs
+++ b/tests/bigint.rs
@@ -415,6 +415,13 @@ fn test_convert_f32() {
         b <<= 1;
     }
 
+    // test correct ties-to-even rounding
+    let weird: i128 = (1i128 << 100) + (1i128 << (100 - f32::MANTISSA_DIGITS));
+    assert_ne!(weird as f32, (weird + 1) as f32);
+
+    assert_eq!(BigInt::from(weird).to_f32(), Some(weird as f32));
+    assert_eq!(BigInt::from(weird + 1).to_f32(), Some((weird + 1) as f32));
+
     // rounding
     assert_eq!(
         BigInt::from_f32(-f32::consts::PI),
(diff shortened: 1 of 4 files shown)


dramforever:
I'm not quite sure how to characterize this.

- Is this a bug? It's certainly pretty surprising.
- Should I add this to the changelog? Is this a breaking change?

cuviper:
In general, the goal is to match the behavior of primitive types, at least so far as that behavior does not depend on limited bit width. So yes, having an example of different `i128` behavior makes this look like a bug we should fix.

Breaking changes are fuzzy when it comes to behavior -- I think this is a justified change, but yes we should mention it in release notes. There's semi-related precedent in how Rust 1.67 fixed rounding in https://github.com/rust-lang/rust/pull/102935/.

tests/bigint.rs
```diff
@@ -421,6 +421,13 @@ fn test_convert_f32() {
         b <<= 1;
     }
 
+    // test correct ties-to-even rounding
+    let weird: i128 = (1i128 << 100) + (1i128 << (100 - 24));
```
cuviper, line 425: I'd rather see `MANTISSA_DIGITS` used instead of magic numbers 24 or 53 -- but that's not just you, I see the existing tests already abuse that.

Maybe we could also add tests in `ci/big_quickcheck`?

dramforever:
ooh that's not quite right, i'll take a look

tests/bigint.rs
dramforever: Changed to use `MANTISSA_DIGITS` for the new tests I've added.

Added a quickcheck test. I had to make my own `Arbitrary` to increase the chance of it failing for the old rounding. A bit ugly though... What do you think?

cuviper:
Thanks! I just rebased this without changes to get CI running.

bors r+

PR #274, opened 2023-07-06
Add true_division: divides 2 big integers and returns a f64

twiby:
Sometimes one would to divide 2 big integers to get a f64. Most of the time, one could convert them to f64 first and then dividing, but in some cases one can lose precision bits (and if they are really big they won't fit in a f64 at all, even though their quotient might).

This is achieved by shifting the rhs to be 53 bits longer than the lhs. The regular integer division is then guaranteed to be 53 or 54 bits long. Then final result packs this quotient as the mantissa, with the exponent.

I'm completely open to renaming/suggestions about how to make this better ! As a matter of fact, this might not be useful at all (I personally have not stumble upon a credible use case), but while building my own big integer library for fun (at github.com:twiby/bigint), I added it by symmetry with Python's `__truediv__`.

In fact, I've never contributed to open_source projects before, and I hope I'm not doing anything wrong in one way or another !

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -24,6 +24,7 @@ mod addition;
 mod division;
 mod multiplication;
 mod subtraction;
+mod true_division;
 
 mod bits;
 mod convert;
--- a/src/bigint/true_division.rs
+++ b/src/bigint/true_division.rs
@@ -1,0 +1,12 @@
+use crate::{BigInt, Sign::*};
+
+impl BigInt {
+    pub fn true_div(&self, other: &BigInt) -> f64 {
+        let d = self.data.true_div(&other.data);
+        match (self.sign, other.sign) {
+            (Plus, Plus) | (NoSign, Plus) | (Minus, Minus) => d,
+            (Plus, Minus) | (NoSign, Minus) | (Minus, Plus) => -d,
+            (_, NoSign) => unreachable!(),
+        }
+    }
+}
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -17,6 +17,7 @@ mod addition;
 mod division;
 mod multiplication;
 mod subtraction;
+mod true_division;
 
 mod bits;
 mod convert;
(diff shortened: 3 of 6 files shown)


cuviper:
I do think this is a reasonable thing to want, although I have quibbles with the "true" name, given how lossy floating point numbers are. The truly accurate representation is a rational number, and in fact `num-rational` does support [`to_f64`](https://docs.rs/num-rational/latest/num_rational/struct.Ratio.html#method.to_f64).

Does that meet your needs? (although you also said you have no use case...)

twiby:
Hi!
No problem with renaming, maybe `float_division` ? 

I tried to use num-rational for this purpose but it was not ideal:
 - The API returns a `Option<f64>::None` instead of a nan, which is very Rusty, but not a pure floating point style result in my opinion.
 - The `num-rational` crate actually imports the `num-bigint` crate, so using it to create a wrapper for this feature in `num-bigint` creates import issues (`Ration` wants the BigInt it itself included, not the current ones). Also a user wanting only this feature would be annoyed to rewrite their import structure.

One avenue could be to put the original mechanism here, and let `num-rational` import it with its own signature (that means a second PR over there). In that case I would use the original implementation (which is the same principle as mine).

What do you think ?

cuviper:
> * The API returns a `Option<f64>::None` instead of a nan, which is very Rusty, but not a pure floating point style result in my opinion.

You could use `.unwrap_or(f64::NAN)`.

> * The `num-rational` crate actually imports the `num-bigint` crate,

Right, we can't have a circular dependency.

> One avenue could be to put the original mechanism here, and let `num-rational` import it with its own signature (that means a second PR over there).

That implementation is also used when its `num-bigint` dependency is not enabled, just using `i128` instead. And even with bigints, it attempts to use `i128` first just for performance. So, I'm not sure there's a great way to separate this without just duplicating the code. :/

twiby:
Ok.
Since `num-bigint` and `num-rational` are in the same ecosystem, I think it would be kind of crazy to duplicate this code. All in all, this pr makes the feature a little more accessible, but is might not be worth it. I'm ok to close this if you agree.

Issue #169, opened 2020-10-09
Algorithm improvements for very large numbers


tczajka:
I am planning to implement a number algorithm improvements for large numbers. Since I am new to the codebase, I would appreciate any feedback or advice.

1. Multiplication currently uses Toom-3, which is O(n^1.46) for n BigDigits. This can be improved to O(n log n) using fast fourier transform (on complex numbers with sufficient precision) or number theoretic transforms (using modular arithmetic on multiple prime numbers), or slightly worse O(n log n log log n) using the Schönhage–Strassen algorithm. I am leaning towards the latter because it's only slightly slower asymptotically, but simpler, so I suspect it might actually be faster for practical sizes.

2. DIvision uses an O(n^2) algorithm, can be improved to same as multiplication O(M(n)).

3. Radix conversion for non-powers of 2 (especially 10) is O(n^2), can be improved to O(M(n) log n).

4. Sqrt is O(M(n) log n), can be improved to O(M(n)).

5. Gcd is O(n^2), can be improved to O(M(n) log n).

6. Unbalanced operations (e.g. 1000 digits x 1000000 digits) can be optimized.

These will require more complex algorithms, so it will probably makes sense to split algorithms into separate modules for multiplication, division, etc.

I think it's worth it, because many things are quadratic in the number of digits now, but after improvements everything will be O(n log^k n).

cuviper:
It would be great to improve the scalability, as long as it's done in a way that doesn't hurt the performance of relatively-small numbers. For example, the current multiplication scales from basic long multiplication to Karatsuba to Toom-3, since constant factors can overwhelm Big-O algorithm scaling. Of course, there may well be possible improvements in small numbers too!

Refactoring the modules is also fine, but please try to do it in incremental steps for ease of review. If you give me a huge PR at once, I'll be less able to deal with it in a timely manner.

SparrowLii:
No doubt it needs to find a suitable threshold to protect small numbers. And It may be a good practice to look at the implementation in GMP.

I did an optimization of big int division in Go before, now doing FFT with Go. I'm happy to help if you need.

SparrowLii:
O(NlogN ) FFT is difficult for integers, as the loss of accuracy cannot be avoided. Number Theoretic Transform is also very difficult, because you can't easily find the right prime number. I think Schönhage–Strassen is a good choice. 
[...]

PR #282, opened 2023-08-28
Implement number theoretic transform for large integer multiplication

byeongkeunahn:
This commit implements [number theoretic transform (NTT)](https://en.wikipedia.org/wiki/Discrete_Fourier_transform_over_a_ring#Number-theoretic_transform) for large integer multiplication (issue #169).

* To simplify implementation the [Schönhage–Strassen algorithm](https://en.wikipedia.org/wiki/Sch%C3%B6nhage%E2%80%93Strassen_algorithm) was not used. Instead, three distinct 64-bit primes were carefully chosen to enable NTT up to ~10^18 64bit integers, which allows multiplication up to ~5 x 10^17 64bit integers. Depending on the input length either two or three primes are used, with the latter only used when the inputs consist of at least 2^40 64bit integers. The convolution results modulo primes are merged using the [Chinese Remainder Theorem](https://en.wikipedia.org/wiki/Chinese_remainder_theorem).
* To reduce padding and the number of cycles for NTT, multiple radices are used (radix-2, radix-3, radix-4, radix-5, radix-6). Radix-8 is desirable but actually slower, presumably due to register spill. Although manual SIMD coding may alleviate this issue, it was not used (i) for maximum portability and (ii) since 64bit SIMD multiply is not widely available even on x86/x64 platforms (AVX512). The prime moduli are carefully chosen to support these radices.
  * The NTT length is selected by exhaustively evaluating cost estimates for all allowed lengths within a factor of two.
* Single-word (u64) [Montgomery reduction](https://en.algorithmica.org/hpc/number-theory/montgomery/) is used for fast modular multiplication.
* 32bit digits are supported by repacking the digits into u64, running the u64 algorithm, and converting back to u32. This results in 32bit builds being about 3-5x slower compared to 64bit builds, which, however, still is an improvement upon the existing algorithms.
* Unbalanced multiplication is enabled when the cost estimates are favorable.
* Based on experimentation, the following thresholds are chosen:
  * For u64 digits, switch to NTT if the shorter integer has at least 512 digits.
  * For u32 digits, switch to NTT if the shorter integer has at least 2,048 digits.
* The three primes are as follows:
  * P1 = 10_237_243_632_176_332_801, Max NTT length = 2^24 * 3^20 * 5^2 = 1_462_463_376_025_190_400
  * P2 = 13_649_658_176_235_110_401, Max NTT length = 2^26 * 3^19 * 5^2 = 1_949_951_168_033_587_200
  * P3 = 14_259_017_916_245_606_401, Max NTT length = 2^22 * 3^21 * 5^2 = 1_096_847_532_018_892_800
  * P1 and P2 are used for the two-prime NTT, whereas the three-prime NTT uses all three.
* The following MIT-licensed projects are used as reference:
  * [http://wwwa.pikara.ne.jp/okojisan/otfft-en/index.html](http://wwwa.pikara.ne.jp/okojisan/otfft-en/index.html)
  * [https://github.com/Bubbler-4/cg263748](https://github.com/Bubbler-4/cg263748)

On Ryzen 7 2700X, 64bit, it takes about 15ms for 2.7Mbits x 2.7Mbits and 170ms for 27Mbits x 27Mbits multiplication. This seems comparable to GMP 6.2.1.

benchmark-20230829

### DIFF
--- a/src/biguint/multiplication.rs
+++ b/src/biguint/multiplication.rs
@@ -420,7 +420,7 @@ fn mac3(mut acc: &mut [BigDigit], mut b: &[BigDigit], mut c: &[BigDigit]) {
         // NTT multiplies two integers by computing the convolution of the arrays
         // modulo a prime. Since the result may exceed the prime, we use two or three
         // distinct primes and combine the results using the Chinese Remainder
-        // Theroem (CRT).
+        // Theorem (CRT).
         ntt::mac3(acc, b, c);
     }
 }
--- a/src/biguint/multiplication.rs
+++ b/src/biguint/multiplication.rs
@@ -11,6 +11,8 @@ use core::iter::Product;
 use core::ops::{Mul, MulAssign};
 use num_traits::{CheckedMul, FromPrimitive, Zero};
 
+use super::ntt;
+
 #[inline]
 pub(super) fn mac_with_carry(
     a: BigDigit,
@@ -93,8 +95,9 @@ fn mac3(mut acc: &mut [BigDigit], mut b: &[BigDigit], mut c: &[BigDigit]) {
     // - If y is at least least twice as long as x, split using Half-Karatsuba.
     // - Next we use Karatsuba multiplication (Toom-2), which we have optimized
     //   to avoid unnecessary allocations for intermediate values.
-    // - For the largest inputs we use Toom-3, which better optimizes the
+    // - Next we use Toom-3, which better optimizes the
     //   number of operations, but uses more temporary allocations.
+    // - For the largest inputs we use number-theoretic transform (NTT).
     //
     // The thresholds are somewhat arbitrary, chosen by evaluating the results
     // of `cargo bench --bench bigint multiply`.
@@ -278,7 +281,11 @@ fn mac3(mut acc: &mut [BigDigit], mut b: &[BigDigit], mut c: &[BigDigit]) {
             }
             NoSign => (),
         }
-    } else {
+    } else if x.len() <= {
+        cfg_32!(let ntt_threshold = 2048;);
+        cfg_64!(let ntt_threshold = 512;);
+        ntt_threshold
+    } {
         // Toom-3 multiplication:
         //
         // Toom-3 is like Karatsuba above, but dividing the inputs into three parts.
@@ -407,6 +414,14 @@ fn mac3(mut acc: &mut [BigDigit], mut b: &[BigDigit], mut c: &[BigDigit]) {
                 NoSign => {}
             }
         }
+    } else {
+        // Number-theoretic transform (NTT) multiplication:
+        //
+        // NTT multiplies two integers by computing the convolution of the arrays
+        // modulo a prime. Since the result may exceed the prime, we use two or three
+        // distinct primes and combine the results using the Chinese Remainder
+        // Theroem (CRT).
+        ntt::mac3(acc, b, c);
     }
 }
 
--- a/src/biguint/multiplication.rs
+++ b/src/biguint/multiplication.rs
@@ -281,11 +281,7 @@ fn mac3(mut acc: &mut [BigDigit], mut b: &[BigDigit], mut c: &[BigDigit]) {
             }
             NoSign => (),
         }
-    } else if x.len() <= {
-        cfg_32!(let ntt_threshold = 2048;);
-        cfg_64!(let ntt_threshold = 512;);
-        ntt_threshold
-    } {
+    } else if x.len() <= 512 {
         // Toom-3 multiplication:
         //
         // Toom-3 is like Karatsuba above, but dividing the inputs into three parts.
--- a/src/biguint/ntt.rs
+++ b/src/biguint/ntt.rs
@@ -1,11 +1,4 @@
 #![forbid(unsafe_code)]
-#![allow(clippy::cast_sign_loss)]
-#![allow(clippy::cast_lossless)]
-#![allow(clippy::cast_possible_truncation)]
-#![allow(clippy::many_single_char_names)]
-#![allow(clippy::needless_range_loop)]
-#![allow(clippy::too_many_arguments)]
-#![allow(clippy::similar_names)]
 
 use super::Vec;
 
@@ -530,6 +523,7 @@ fn ntt4_single_block<const P: u64, const INV: bool, const TWIDDLE: bool>(px: &mu
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 const fn ntt5_kernel<const P: u64, const INV: bool, const TWIDDLE: bool>(
     w1: u64,
     w2: u64,
@@ -591,6 +585,7 @@ fn ntt5_single_block<const P: u64, const INV: bool, const TWIDDLE: bool>(px: &mu
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 const fn ntt6_kernel<const P: u64, const INV: bool, const TWIDDLE: bool>(
     w1: u64,
     w2: u64,
--- a/src/biguint/ntt.rs
+++ b/src/biguint/ntt.rs
@@ -418,10 +418,10 @@ impl<const P: u64, const INV: bool> NttKernelImpl<P, INV> {
 }
 
 const fn ntt2_kernel<const P: u64, const INV: bool, const TWIDDLE: bool>(
-    w1: u64,
     a: u64,
     mut b: u64,
-) -> (u64, u64) {
+    w1: u64,
+) -> [u64; 2] {
     if !INV && TWIDDLE {
         b = Arith::<P>::mmulmod(w1, b);
     }
@@ -430,7 +430,7 @@ const fn ntt2_kernel<const P: u64, const INV: bool, const TWIDDLE: bool>(
     if INV && TWIDDLE {
         out1 = Arith::<P>::mmulmod(w1, out1);
     }
-    (out0, out1)
+    [out0, out1]
 }
 
 fn ntt2_single_block<const P: u64, const INV: bool, const TWIDDLE: bool>(px: &mut [u64], w1: u64) {
@@ -438,17 +438,16 @@ fn ntt2_single_block<const P: u64, const INV: bool, const TWIDDLE: bool>(px: &mu
     let s1 = px.len() / 2;
     let (a, b) = px.split_at_mut(s1);
     for (a, b) in a.iter_mut().zip(b) {
-        (*a, *b) = ntt2_kernel::<P, INV, TWIDDLE>(w1, *a, *b);
+        [*a, *b] = ntt2_kernel::<P, INV, TWIDDLE>(*a, *b, w1);
     }
 }
 
 const fn ntt3_kernel<const P: u64, const INV: bool, const TWIDDLE: bool>(
-    w1: u64,
-    w2: u64,
     a: u64,
     mut b: u64,
     mut c: u64,
-) -> (u64, u64, u64) {
+    [w1, w2]: [u64; 2],
+) -> [u64; 3] {
     if !INV && TWIDDLE {
         b = Arith::<P>::mmulmod(w1, b);
         c = Arith::<P>::mmulmod(w2, c);
@@ -461,29 +460,31 @@ const fn ntt3_kernel<const P: u64, const INV: bool, const TWIDDLE: bool>(
         out1 = Arith::<P>::mmulmod(w1, out1);
         out2 = Arith::<P>::mmulmod(w2, out2);
     }
-    (out0, out1, out2)
+    [out0, out1, out2]
 }
 
 fn ntt3_single_block<const P: u64, const INV: bool, const TWIDDLE: bool>(px: &mut [u64], w1: u64) {
-    let w1 = if TWIDDLE { w1 } else { 0 };
-    let w2 = Arith::<P>::mmulmod(w1, w1);
+    let w = if TWIDDLE {
+        let w2 = Arith::<P>::mmulmod(w1, w1);
+        [w1, w2]
+    } else {
+        [0; 2]
+    };
     let s1 = px.len() / 3;
     let (a, rest) = px.split_at_mut(s1);
     let (b, c) = rest.split_at_mut(s1);
     for ((a, b), c) in a.iter_mut().zip(b).zip(c) {
-        (*a, *b, *c) = ntt3_kernel::<P, INV, TWIDDLE>(w1, w2, *a, *b, *c);
+        [*a, *b, *c] = ntt3_kernel::<P, INV, TWIDDLE>(*a, *b, *c, w);
     }
 }
 
 const fn ntt4_kernel<const P: u64, const INV: bool, const TWIDDLE: bool>(
-    w1: u64,
-    w2: u64,
-    w3: u64,
     a: u64,
     mut b: u64,
     mut c: u64,
     mut d: u64,
-) -> (u64, u64, u64, u64) {
+    [w1, w2, w3]: [u64; 3],
+) -> [u64; 4] {
     if !INV && TWIDDLE {
         b = Arith::<P>::mmulmod(w1, b);
         c = Arith::<P>::mmulmod(w2, c);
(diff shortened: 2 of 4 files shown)


cuviper:
I want to first acknowledge this with thanks -- it's quite impressive to be near GMP performance!

However to set expectations: this is also a large PR, and I will need some studying to understand what's going on, so it may take me a while to find time to review this.

byeongkeunahn:
Thanks. I'd like to note a few changes in the latest commit:

* The updated code uses decimation-in-frequency (DIF) in forward transform and decimation-in-time (DIT) in inverse transform.
* The prime numbers are replaced with a bigger primes to enable tighter bit-packing in some cases that occur periodically.
  * P1 = 14_259_017_916_245_606_401, Max NTT length = 2^22 * 3^21 * 5^2 = 1_096_847_532_018_892_800
  * P2 = 17_984_575_660_032_000_001, Max NTT length = 2^19 * 3^17 * 5^6 = 1_057_916_215_296_000_000
  * P3 = 17_995_154_822_184_960_001, Max NTT length = 2^17 * 3^22 * 5^4 = 2_570_736_403_169_280_000
  * P2 and P3 are used for the two-prime NTT, whereas the three-prime NTT uses all three.

HKalbasi:
I ran benchmark fib_hex 100m from https://github.com/tczajka/bigint-benchmark-rs on this PR and it made num-bigint twice faster than malachite, slightly faster than gmp and 12x faster than itself.

cuviper:
Obviously the performance here is impressive, as I said before. However, as I finally started to try reviewing this, I have a few high level objections.

- The long commit history is not something I would want to merge. If there are discrete logical steps in implementing this feature, we can preserve some of that, but not a bunch of "update", "improve", and merges.
- The Montgomery implementation should share `src/biguint/monty.rs`, extending that if needed.
- So much `unsafe` code is uncharacteristic for this crate, and I would much rather stick to safe code as much as possible, even if that means sacrificing some performance. Case in point, miri finds UB here:

```
test test_mul_overflow_2 ... error: Undefined Behavior: attempting a write access using <5072044> at alloc1162269[0x0], but that tag does not exist in the borrow stack for this location
   --> src/biguint/ntt.rs:230:13
    |
230 |             *out.add(i) = Arith::<P>::mreduce(v);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this error occurs as part of an access at alloc1162269[0x0..0x8]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <5072044> was created by a SharedReadWrite retag at offsets [0x30..0x4830]
   --> src/biguint/ntt.rs:518:27
    |
518 |         conv_base::<P>(g, (&mut x[i..]).as_mut_ptr(), (&mut y[i..]).as_mut_ptr(), tf_current);
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: this is on thread `test_mul_overfl`
    = note: stack backtrace:
            0: num_bigint::biguint::ntt::conv_base::<17984575660032000001>
                at src/biguint/ntt.rs:230:13: 230:49
            1: num_bigint::biguint::ntt::conv::<17984575660032000001>
                at src/biguint/ntt.rs:518:9: 518:94
            2: num_bigint::biguint::ntt::mac3_two_primes
                at src/biguint/ntt.rs:604:5: 604:98
            3: num_bigint::biguint::ntt::mac3_u64
                at src/biguint/ntt.rs:760:9: 760:41
            4: num_bigint::biguint::ntt::mac3
                at src/biguint/ntt.rs:772:5: 772:24
            5: num_bigint::biguint::multiplication::mac3
                at src/biguint/multiplication.rs:358:9: 358:29
            6: num_bigint::biguint::multiplication::mul3
                at src/biguint/multiplication.rs:366:5: 366:31
            7: num_bigint::biguint::multiplication::<impl std::ops::Mul for &num_bigint::BigUint>::mul
                at src/biguint/multiplication.rs:428:31: 428:41
            8: test_mul_overflow_2
                at tests/biguint.rs:1075:20: 1075:27
            9: test_mul_overflow_2::{closure#0}
                at tests/biguint.rs:1068:25: 1068:25
```

(that particular one does pass under tree-borrows though)

byeongkeunahn:
Thanks for your review and comments! I’ll try to address the issues, including removing the unsafe code and cleaning up the commit history, even if that comes at the cost of a small performance regression.

Integrating the Montgomery reduction code with `src/monty.rs` is a bit more complicated, though:

* The NTT code currently uses a 64-bit modulus regardless of the target architecture, whereas `src/monty.rs` uses `BigDigit`. Introducing generics could potentially address this mismatch.
* The Montgomery operations in this PR were hand-optimized by inspecting the LLVM output on Godbolt and benchmarking promising variants. They are tailored specifically to the operations used by the NTT routine, including fused operations and the careful omission of overflow checks where they are mathematically unnecessary. As a result, they serve a somewhat different purpose from the code in `src/monty.rs`, and I’m not yet sure how they could be cleanly integrated there.

cuviper:
We can treat the monty-consolidation as "nice to have". However, any specific tailoring needs comments, lest that work be undone by a later contributor or maintainer. I'm skeptical that those optimizations wouldn't be useful to the other monty use as well though...

byeongkeunahn:
I’ve made the suggested improvements:

* Cleaned up the commit history.
  The original branch remains available [here](https://github.com/byeongkeunahn/num-bigint/tree/master-backup) for reference.
* Rebased the branch onto the latest `main`.
* Replaced the unsafe routines with safe equivalents. Benchmarks showed no performance regression.
* Added comments explaining why the NTT routine requires its own Montgomery arithmetic routines.

I also made a small improvement to the planner based on exhaustive Criterion.rs benchmarks covering operand sizes of up to 300 million bits each. This reduced execution time by 1.2% on average.

Some of this work was completed with assistance from ChatGPT Codex.

Thanks again for the detailed review. Please let me know if there are any remaining issues.

src/biguint/ntt.rs
```diff
@@ -0,0 +1,1091 @@
+#![forbid(unsafe_code)]
+#![allow(clippy::cast_sign_loss)]
+#![allow(clippy::cast_lossless)]
+#![allow(clippy::cast_possible_truncation)]
+#![allow(clippy::many_single_char_names)]
+#![allow(clippy::needless_range_loop)]
+#![allow(clippy::too_many_arguments)]
+#![allow(clippy::similar_names)]
+
+use crate::biguint::Vec;
```
cuviper, line 10: It's odd to use an absolute path when it's not an intended re-export.
```suggestion
use super::Vec;
```

src/biguint/ntt.rs
```diff
@@ -0,0 +1,1091 @@
...
+        loop {
+            if a == 0 {
+                break (b, c[1], c[3]);
+            }
+            let (q, r) = (b / a, b % a);
+            (a, b) = (r, a);
+            c = [c[1] - q * c[0], c[0], c[3] - q * c[2], c[2]];
+        }
+    }
+    // Modular inverse: a^-1 mod modulus
```
cuviper, line 32: Style nit here and throughout: please add a blank line between functions/consts/etc.

Even within code blocks, if there's a comment explaining the following code, then also add a blank line before. (except if it's the first part at a new indentation level)

src/biguint/ntt.rs
```diff
@@ -0,0 +1,1091 @@
...
+    pub last_radix: usize,
+    pub s_list: Vec<(usize, usize)>,
+}
+impl NttPlan {
+    fn build<const P: u64>(min_len: usize) -> Self {
+        assert!(min_len as u64 <= Arith::<P>::MAX_NTT_LEN);
+        let (mut len_max, mut len_max_cost, mut g) = (usize::MAX, usize::MAX, 1);
+        for m7 in 0..=1 {
+            for m5 in 0..=Arith::<P>::factors(5) {
+                for m3 in 0..=Arith::<P>::factors(3) {
```
cuviper, line 208: Might it be worth forcing const-evaluation on these? i.e. using a `const { ... }` block, or perhaps adding `Arith::FACTORS_5` etc.

src/biguint/ntt.rs
```diff
@@ -0,0 +1,1091 @@
...
+                    let len = 7u64.pow(m7) * 5u64.pow(m5) * 3u64.pow(m3);
+                    if len >= 2 * min_len as u64 {
+                        break;
+                    }
+                    let (mut len, mut m2) = (len as usize, 0);
+                    while len < min_len && m2 < Arith::<P>::factors(2) {
```
cuviper, line 214: Also FWIW, `factors(2)` is just `(P-1).trailing_zeros()` -- but it doesn't really matter either way if we're doing it at compile time.

src/biguint/ntt.rs
```diff
@@ -0,0 +1,1091 @@
...
+                        len *= 2;
+                        m2 += 1;
+                    }
+                    if len >= min_len && len < len_max_cost {
+                        let (mut tmp, mut cost) = (len, 0);
+                        let mut g_new = 1;
+
+                        // Length-dependent weights for cost estimation.
```
cuviper, line 222: These weights and cost formulas look like a whole lot of "magic" constants. I suppose they come from your PR statement:
> The NTT length is selected by exhaustively evaluating cost estimates for all allowed lengths within a factor of two.

It would be good to have comments directly in the code to explain this, ideally with some instructions how one might reevaluate these costs. If there's any third-party source as well, please link to that.

src/biguint/multiplication.rs
```diff
@@ -407,6 +414,14 @@ fn mac3(mut acc: &mut [BigDigit], mut b: &[BigDigit], mut c: &[BigDigit]) {
                 NoSign => {}
             }
         }
+    } else {
+        // Number-theoretic transform (NTT) multiplication:
+        //
+        // NTT multiplies two integers by computing the convolution of the arrays
+        // modulo a prime. Since the result may exceed the prime, we use two or three
+        // distinct primes and combine the results using the Chinese Remainder
+        // Theroem (CRT).
```
cuviper, line 423:
```suggestion
        // Theorem (CRT).
```

src/biguint/ntt.rs
```diff
@@ -0,0 +1,1091 @@
...
+    pub g: usize, // g: size of the base case
+    pub m: usize, // m divides Arith::<P>::MAX_NTT_LEN
+    pub last_radix: usize,
+    pub s_list: Vec<(usize, usize)>,
+}
+impl NttPlan {
+    fn build<const P: u64>(min_len: usize) -> Self {
+        assert!(min_len as u64 <= Arith::<P>::MAX_NTT_LEN);
+        let (mut len_max, mut len_max_cost, mut g) = (usize::MAX, usize::MAX, 1);
+        for m7 in 0..=1 {
```
cuviper, line 206: Can you explain (and add comments) why we're checking 7 this way, regardless of factors?

src/biguint/ntt.rs
```diff
@@ -0,0 +1,1091 @@
...
+// subtractive reduction remains const-evaluable, handles primes close to
+// 2^64 without a 129-bit addition, and supports the fused and delayed
+// reductions required by the NTT kernels.
+struct Arith<const P: u64> {}
+impl<const P: u64> Arith<P> {
+    const R: u64 = ((1u128 << 64) % P as u128) as u64; // 2^64 mod P
+    const R2: u64 = (Self::R as u128 * Self::R as u128 % P as u128) as u64; // R^2 mod P
+    const PINV: u64 = arith::invmod_2pow64(P); // P^-1 mod 2^64
+    const MAX_NTT_LEN: u64 =
+        2u64.pow(Self::factors(2)) * 3u64.pow(Self::factors(3)) * 5u64.pow(Self::factors(5));
```
cuviper, line 72: We expect that `P-1` is *nearly* 5-smooth, right? So the remaining (`(P - 1) / MAX_NTT_LEN`) factor should be small -- maybe that's worth a const-assert here? What's are the actual properties required in that?

I see that the chosen primes have additional `P-1` factors 13, 17, and 7, but we should make sure to explain the significance in case someone wants to try other values.

src/biguint/ntt.rs
```diff
@@ -0,0 +1,1091 @@
+#![forbid(unsafe_code)]
+#![allow(clippy::cast_sign_loss)]
+#![allow(clippy::cast_lossless)]
+#![allow(clippy::cast_possible_truncation)]
+#![allow(clippy::many_single_char_names)]
+#![allow(clippy::needless_range_loop)]
+#![allow(clippy::too_many_arguments)]
+#![allow(clippy::similar_names)]
```
cuviper, line 8: I'm not interested in maintaining a bunch of _pedantic_ clippy suppressions. AFAICS, the only lint here that we hit by default is `too_many_arguments`, and that's only on `ntt5_kernel` and `ntt6_kernel`, so let's just `allow` specifically on those.

src/biguint/ntt.rs
```diff
@@ -0,0 +1,1091 @@
...
+    x: &mut [u64],
+    xlen: usize,
+    y: &mut [u64],
+    ylen: usize,
+    mut mult: u64,
+) {
+    assert!(!x.is_empty() && x.len() == y.len());
+    let (_n, g, m, last_radix) = (plan.n, plan.g, plan.m, plan.last_radix as u64);
+
+    /* multiply by a constant in advance */
```
cuviper, line 714: Another style nit: please use `//` comments throughout, rather than `/* */`

src/biguint/ntt.rs
```diff
@@ -0,0 +1,1091 @@
...
+const P1P2INV_R_MOD_P3: u64 = Arith::<P3>::mmulmod(
+    Arith::<P3>::R2,
+    arith::invmod((P1 as u128 * P2 as u128 % P3 as u128) as u64, P3),
+);
+const P1_R_MOD_P3: u64 = Arith::<P3>::mmulmod(Arith::<P3>::R2, P1);
+const P1P2_LO: u64 = (P1 as u128 * P2 as u128) as u64;
+const P1P2_HI: u64 = ((P1 as u128 * P2 as u128) >> 64) as u64;
+
+// Propagates carry from the beginning to the end of acc,
+//   and returns the resulting carry if it is nonzero.
```
cuviper, line 803:
```suggestion
// Propagates carry from the beginning to the end of acc,
// and returns the resulting carry if it is nonzero.
```

src/biguint/ntt.rs
```diff
@@ -0,0 +1,1091 @@
...
+            if exponent % 2 > 0 {
+                cur = Self::mmulmod(cur, base);
+            }
+            exponent /= 2;
+            base = Self::mmulmod(base, base);
+        }
+        cur
+    }
+    // Computes c as u128 * mreduce(v) as u128,
+    //   using d: u64 = mmulmod(P-1, c).
```
cuviper, line 145:
```suggestion
    // Computes c as u128 * mreduce(v) as u128, using d: u64 = mmulmod(P-1, c).
    //
```

src/biguint/ntt.rs
```diff
@@ -0,0 +1,1091 @@
...
+            acc[j] = w;
+            carry_acc = u64::from(overflow1 || overflow2);
+
+            /* roll-over */
+            (j, p) = (j + 1, p - 64);
+            bitbuf = out >> (bits - p);
+        }
+    }
+    // Process remaining carries. The addition carry_acc + bitbuf should not overflow
+    //   since bitbuf is underfilled and carry_acc is always 0 or 1.
```
cuviper, line 911:
```suggestion
    // Process remaining carries. The addition carry_acc + bitbuf should not overflow
    // since bitbuf is underfilled and carry_acc is always 0 or 1.
```

cuviper (changes requested):

src/biguint/ntt.rs
cuviper: For too-many-args, it might make sense to switch those to arrays, e.g. 
```rust
const fn ntt5_kernel<const P: u64, const INV: bool, const TWIDDLE: bool>(
    [w1, w2, w3, w4]: [u64; 4],
    [a, mut b, mut c, mut d, mut e]: [u64; 5],
) -> [u64; 5] {
```

... but I haven't checked whether that affects benchmarks.

src/biguint/ntt.rs
cuviper: I've chosen to remove this for now.

cuviper (approved):
@byeongkeunahn I've made the remaining changes myself. You (or anyone else) are welcome to revisit this any time to confirm that I didn't mess anything up or misrepresent any comments, etc. With that, I think this is ready to merge!

byeongkeunahn:
Hi, thanks for the changes! I’ve been reviewing this and verified via benchmarks that using fixed-length slices does not affect performance; I also checked that the generated assembly is unchanged. I’ve also been examining the planner and gathering additional data.

For prime selection, we should balance two considerations: the modulus should be large to increase the packing density, while also having as many factors of 2, 3, and 5 as possible. Currently, the primes are selected by exhaustive search so that they have sufficient factors of 2, 3, and 5 while remaining close to 2^64.

The m7 check was for the base case, which is handled by naive quadratic multiplication, so we can include at most one factor of 7 in the length. This reduces the amount of padding required for certain input lengths.

Besides this, I’ve carefully checked the planner scheme and found that it can be somewhat simplified while retaining the same planning accuracy. I’ve been preparing those changes, and I should have communicated this earlier—sorry about that.

----

The updated planner is available [here](https://github.com/byeongkeunahn/num-bigint/blob/rework-2/src/biguint/ntt.rs). Probably I could open a new PR.

PR #286, opened 2023-09-27
WIP: Moving internals from Vec to Cow (Starting point for discussion)

### LINKED ISSUE #285: Ability to create const BigUint etc.
Instead of a `Vec<u32>` using a `Cow<'_, [u32]>` would enable people to create a const BigUint by just passing in a &'static [u32].

Due to the nature of bigints I tend to see people creating a bunch of constant values at runtime.

Sure, it's not really that expensive, but it would be nice.

junderw:
Fixes #285 

This will allow for const creation of BigUint etc.

I started off by just doing BigUint to give an idea of what it would take. Please use this as a starting off point for discussion.

Right now I'm not looking for optimization, but just the bare minimum.

If you have a static borrow and something in the original code calls a mut method, a copy will be made, regardless of whether the mut method would have actually changed the contents of the Vec or not.

### DIFF
--- a/src/biguint.rs
+++ b/src/biguint.rs
@@ -1,5 +1,5 @@
 use crate::big_digit::{self, BigDigit};
-use crate::std_alloc::{String, Vec};
+use crate::std_alloc::{Cow, String, Vec};
 
 use core::cmp;
 use core::cmp::Ordering;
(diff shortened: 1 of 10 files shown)


cuviper:
I did try to discuss this on the issue #285, but you never responded. I'm going with my enum approach.

PR #294, opened 2024-03-04
Avoid cloning inputs in sub and add

jonas-lj:
Addition and subtraction with references currently clones inputs to use in the case that the other input is zero. For all other cases, the cloned value is not used.

This PR ensures that the clone is done lazily to avoid the performance penalty of the unnecessary clone.

The PR also uses a function instead of a macro, but this could be easily reverted.


cuviper:
I may be missing something, but where are there unnecessary clones? The macro expressions like `$a_owned` only get expanded and evaluated in the branches that actually use them.

jonas-lj:
Ah, I think you're right. It's because the macro input parameters are expressions. Sorry about that, and thanks for the feedback!

PR #297, opened 2024-04-30
feat: Implement Step trait for BigInt and BigUint

### LINKED ISSUE #68: implement Trait std::iter::Step for BigInt
Maybe it will be more convenient to write the following code  after adding trait `Step` for `BigInt`: 
```rust
for i in BigInt::from(1)..BigInt::from(1000) {
    // do something
}
```

thecaralice:
Implements unstable `Step` trait under a feature gate.

Closes #68.


cuviper:
Hi - thank you, but I am not interested in maintaining unstable features here. You could possibly still do this on your own with a wrapper type, but as I said in #68, it will probably be faster to use a primitive integer type and convert to bigints.

PR #45, opened 2018-05-16
Document Toom3 algorithm

tiehuis:
Happened to be going through toom3 so thought I'd type this up.

PR #304, opened 2024-05-09
Fixed incorrect formulas for coefficients.

ralphtandetzky:
This concerns only the comments for big integer multiplication where the formulas for the coefficients w0..w4 are explicitly listed. These were not completely correct. Please compare the formulas with https://en.wikipedia.org/wiki/Toom%E2%80%93Cook_multiplication#Interpolation.


cuviper:
Huh, those errors existed ever since the comments were added in #45. Thanks for the fix!

PR #309, opened 2024-05-28
style: use strip_prefix instead of manual starts_with/slice

yhx-12243:
`strip_prefix` is stable [since 1.45.0](https://doc.rust-lang.org/nightly/std/primitive.str.html#method.strip_prefix), which is less than the MSRV (1.60) of this crate, and it generally [generates better code](https://rust.godbolt.org/z/nfcsv4Eb8).

### DIFF
--- a/src/bigint/convert.rs
+++ b/src/bigint/convert.rs
@@ -25,8 +25,7 @@ impl Num for BigInt {
     /// Creates and initializes a [`BigInt`].
     #[inline]
     fn from_str_radix(mut s: &str, radix: u32) -> Result<BigInt, ParseBigIntError> {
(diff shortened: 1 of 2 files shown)


cuviper:
I'm surprised clippy didn't see this, but maybe it only lints if there's a manual `len()` too.
https://rust-lang.github.io/rust-clippy/master/#/manual_strip

Thanks!

yhx-12243:
> I'm surprised clippy didn't see this, but maybe it only lints if there's a manual `len()` too. https://rust-lang.github.io/rust-clippy/master/#/manual_strip
> 
> Thanks!

Yes I've already proposed at rust-lang/rust-clippy#12863. 👀

PR #310, opened 2024-06-01
added BigURational as a "superset"

d3crvpt3d:
i know this isnt really sorted how you like it,
but i added a unsigned rational type for handling BigURationals, i'll try to add BigRationals aswell this week

shouldnt throw errors, but i have the plan to maintain this fork


cuviper:
We already have a generic version in https://github.com/rust-num/num-rational

Issue #315, opened 2024-12-13
Convert to base 10 is significantly slower than similar libraries


HKalbasi:
Other libraries like GMP use a divide and conquer algorithm which find a power of 10 with size near half of the digits of the number, use that as the base and convert the original number to a 2 digit number in that base, convert each digit recursively to base 10 and finally concat the results. Assuming that we have a `O(nlogn)` multiplication routine, this algorithm would be `O(nlog^2 n)` and would be faster than the current `O(n sqrt n)` algorithm.

cuviper:
The current algorithm came in #216, and I mentioned then:
> I honestly expect there's still room for improvement,

Help is welcome!

HKalbasi:
I'm interested in helping, `num-bigint` is the default choice for arbitrary precision integers in Rust so I think it is important to make it fast.

The first step is to merge #282 or something like that since the above algorithm requires a fast multiplication algorithm.

vxgmichel:
For the record, I ran the comparison from [my previous comment](https://github.com/rust-num/num-bigint/issues/315#issuecomment-2701210756) with `num-bigint 0.5.1` (now that #316 is merged):
```shell
$ time ./target/release/bigint
0,98s user 0,02s system 99% cpu 0,991 total
$ time PYTHONINTMAXSTRDIGITS=0 python -c "str(1 << (1 << 23))"
0,41s user 0,02s system 99% cpu 0,426 total
```
It's much better 👍

PR #316, opened 2024-12-16
Use divide and conquer in to_radix_digits

### LINKED ISSUE #323: Performance issue on `BigUint::div_rem`
As I wrote in a [previous comment](https://github.com/rust-num/num-bigint/issues/315#issuecomment-2701210756), I noticed that the performance of `BigUint::div_rem` for large numbers is worse than python `int.__divmod__`.

Consider the following computation, i.e the splitting of `2^2^23` into two digits of base `10**1262611` (`1262611` being half the size of `2^2^23` in decimal digits). It takes 1 second to run using python 3.13:
```shell
$ time python -c "divmod(1<<(1<<23), 10**1262611)"
1,04s user 0,01s system 99% cpu 1,051 total
```

Now consider the same computation using `num-bigint`:
```rust
use num_bigint::BigUint;
use num_integer::Integer;

fn main() {
    let x = BigUint::from(1u32) << (1u32 << 23);
    let w = 1262611; // Half the size of x in decimal digits
    x.div_rem(&BigUint::from(10u32).pow(w));
}
```

It takes more than 4 seconds:
```shell
% time ./target/release/bigint
./target/release/bigint  4,44s user 0,00s system 99% cpu 4,443 total
```

For the record, the `int.__divmod__` implementation is written in pure python:
https://github.com/python/cpython/blob/2904ec2273762df58645a8e245b2281884855b8c/Lib/_pylong.py#L528-L531

Apparently, the complexity of this algorithm is `O(n**1.58)`:
> Its time complexity is O(n**1.58), where n = #bits(a) + #bits(b).

hkalbasi:
This implements the algorithm mentioned in #315

Benchmark:
```
// prev
running 4 tests
test to_str_radix_10      ... bench:       4,242.41 ns/iter (+/- 712.43)
test to_str_radix_10_2    ... bench:      82,360.73 ns/iter (+/- 13,304.91)
test to_str_radix_10_3    ... bench:   3,929,829.90 ns/iter (+/- 514,647.85)
test to_str_radix_10_4    ... bench: 243,146,081.10 ns/iter (+/- 44,213,689.70)
// now
running 4 tests
test to_str_radix_10      ... bench:       4,261.38 ns/iter (+/- 522.98)
test to_str_radix_10_2    ... bench:      83,358.54 ns/iter (+/- 11,170.02)
test to_str_radix_10_3    ... bench:   2,623,301.20 ns/iter (+/- 279,505.89)
test to_str_radix_10_4    ... bench: 195,073,240.50 ns/iter (+/- 17,025,687.08)
```

Currently both grow with `O(n^2)`, to make things algorithmically faster we need a faster multiplication and division algorithm.

### DIFF
--- a/src/biguint/convert.rs
+++ b/src/biguint/convert.rs
@@ -1,7 +1,7 @@
 // This uses stdlib features higher than the MSRV
 #![allow(clippy::manual_range_contains)] // 1.35
 
-use super::{biguint_from_vec, BigUint, IntDigits, ToBigUint};
+use super::{biguint_from_vec, BigUint, ToBigUint};
 
 use super::addition::add2;
 use super::division::{div_rem_digit, FAST_DIV_WIDE};
@@ -16,7 +16,7 @@ use core::cmp::Ordering::{Equal, Greater, Less};
 use core::convert::TryFrom;
 use core::mem;
 use core::str::FromStr;
-use num_integer::{Integer, Roots};
+use num_integer::Integer;
 use num_traits::float::FloatCore;
 use num_traits::{FromPrimitive, Num, PrimInt, ToPrimitive, Zero};
 
@@ -702,7 +702,7 @@ pub(super) fn to_radix_digits_le(u: &BigUint, radix: u32) -> Vec<u8> {
     // Estimate how big the result will be, so we can pre-allocate it.
     let mut res = Vec::with_capacity(radix_digits.to_usize().unwrap_or(0));
 
-    let mut digits = u.clone();
+    let digits = u.clone();
 
     // X86 DIV can quickly divide by a full digit, otherwise we choose a divisor
     // that's suitable for `div_half` to avoid slow `DoubleBigDigit` division.
--- a/src/biguint/convert.rs
+++ b/src/biguint/convert.rs
@@ -1,7 +1,7 @@
 // This uses stdlib features higher than the MSRV
 #![allow(clippy::manual_range_contains)] // 1.35
 
-use super::{biguint_from_vec, BigUint, ToBigUint};
+use super::{biguint_from_vec, BigUint, IntDigits, ToBigUint};
 
 use super::addition::add2;
 use super::division::{div_rem_digit, FAST_DIV_WIDE};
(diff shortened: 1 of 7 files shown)


src/biguint/convert.rs
```diff
@@ -701,34 +701,48 @@ pub(super) fn to_radix_digits_le(u: &BigUint, radix: u32) -> Vec<u8> {
     // The threshold for this was chosen by anecdotal performance measurements to
     // approximate where this starts to make a noticeable difference.
     if digits.data.len() >= 64 {
```
cuviper, line 703: Did you re-evaluate this threshold at all? Notably, it's different than the one you used in `to_radix_digits_le_divide_and_conquer`. Maybe that does make sense since the inner part doesn't have to pay for creating `big_bases`, but I'm not sure.

HKalbasi: These are new results:
```
simple:
test 1009 bit      ... bench:       4,169.26 ns/iter (+/- 470.97)
test 2009 bit    ... bench:      14,735.97 ns/iter (+/- 1,819.63)
test 3009 bit    ... bench:      32,522.20 ns/iter (+/- 2,949.82)
test 4009 bit    ... bench:      56,441.64 ns/iter (+/- 6,354.65)
divide and conquer:
test 1009 bit       ... bench:       5,955.14 ns/iter (+/- 859.07)
test 2009 bit       ... bench:      12,731.82 ns/iter (+/- 1,780.59)
test 3009 bit       ... bench:      18,701.03 ns/iter (+/- 2,284.40)
test 4009 bit       ... bench:      27,605.41 ns/iter (+/- 5,229.87)
```
So probably 2000/64 ~ 32 make sense as new threshold?

> since the inner part doesn't have to pay for creating big_bases, but I'm not sure

If I understand correctly, the main difference in small numbers is that the recursive algorithm loses const propagation for 10. If it wasn't the case, I'd expect some threshold near 8.

cuviper: What exactly did you change for your new results?

HKalbasi: In benchmarks? I changed them to this:
```

#[bench]
fn to_str_radix_10(b: &mut Bencher) {
    to_str_radix_bench(b, 10, 1009);
}

#[bench]
fn to_str_radix_10_2(b: &mut Bencher) {
    to_str_radix_bench(b, 10, 2009);
}

#[bench]
fn to_str_radix_10_3(b: &mut Bencher) {
    to_str_radix_bench(b, 10, 3009);
}

#[bench]
fn to_str_radix_10_4(b: &mut Bencher) {
    to_str_radix_bench(b, 10, 4009);
}
```

And I changed `if digits.data.len() >= 64 {` to `if digits.data.len() >= 1 {` and `    if digits.data.len() >= 1000 {`.

HKalbasi:
I would like to implement [](https://pure.mpg.de/rest/items/item_1819444_4/component/file_2599480/content) for fast division to make to_radix even faster. Would you rather have it in this PR or merge this alone?

HKalbasi:
I implemented the Burnikel Ziegler, and the result is no longer quadratic:

```
running 4 tests
test to_str_radix_10      ... bench:       4,447.95 ns/iter (+/- 504.38)
test to_str_radix_10_2    ... bench:      88,899.42 ns/iter (+/- 14,242.49)
test to_str_radix_10_3    ... bench:   1,968,791.90 ns/iter (+/- 231,474.81)
test to_str_radix_10_4    ... bench:  58,370,762.50 ns/iter (+/- 3,103,897.08)
```
Now the time complexity is `O(n^log3)`. Further improvement needs faster multiplication algorithm.

cuviper:
Sorry I've neglected this -- I will try to review it soon!

cuviper:
I rebased this and applied my own further cleanups. Thanks a lot for your work!

PR #318, opened 2025-01-22
fix clippy lints on tests

bionicles:
breaking up my last PR into smaller chunks as suggested by @cuviper 

this contains only changes from my first commit toward https://github.com/rust-num/num-bigint/pull/317

all changes were only acceptance of suggestions from clippy in the "tests" folder


cuviper:
We need to be careful not to change what tests are doing -- in particular, `test_cmp` is *intending* to test all those weird variations, even if they're not idiomatic.

PR #337, opened 2026-01-02
Update documentation to use Markdown links for `BigUint` and `BigInt` references for consistency.

pRizz:
There are spots in the library that have these links and other spots that don't. I added them everywhere for consistency.

There were a few spots that had to use `super` refs, probably because they were used in macros and in a different module.

I tested the links with `cargo doc` and they work.


cuviper:
Thanks!

PR #322, opened 2025-03-03
Upgrade `rand` to 0.9

divergentdave:
This bumps the crate version, MSRV, and `rand` dependency. I also changed some panics into errors in trait methods that are now fallible, and renamed the methods of `RandBigInt` to align with changes in the `Rng` trait.

PR #338, opened 2026-02-04
Upgrade to `rand` v0.10 and/or v0.9, and split `rand_core`

### LINKED ISSUE #328: Update to rand 0.9.0
The breaking v0.9.0 update of the rand crate has been released since Jan 27, 2025 and your dependecy on the rand crate is still set to v0.8.0. Could you update the dependency to allow interoperability with the newer version of rand?

cuviper:
This now implements `RandBigInt` with only `rand_core::Rng` to allow a
slimmer dependency tree. The other distribution-related types still
require `rand` for its traits though.

Our `gen_big*` methods are also renamed to `random_big*` to match the
change that `rand` had made in 0.9. The old names are deprecated, but
still present, simply forwarding to the new names.

Hat tip to @bionicles and @divergentdave as well for their `rand v0.9` PRs (#317, #322),
and I apologize for never getting around to those. Let's reset that debt by getting
0.10 much sooner!

The crate features are now explicitly versioned, allowing _either_ 0.9 or 0.10, or both.
So we now have `rand_0_9`, `rand_core_0_9`, `rand_0_10`, and `rand_core_0_10`.

`RandBigInt` is also renamed and forked to `BigRng09` and `BigRng010`, so the name
is versioned and better implies that it is an extension of `Rng`. The other distribution
structs are simply shared with trait implementations for either/both versions.


Cargo.toml
```diff
@@ -56,7 +57,12 @@ features = ["i128"]
 
 [dependencies.rand]
 optional = true
-version = "0.8"
+version = "0.10.0-rc.9"
```
cuviper: I'll wait in draft until 0.10.0 is released.

bionicles:
0.10 is out! looks good to me, i copied your homework to debug my fork

[rand 0.10 update guide](https://rust-random.github.io/book/update-0.10.html) if this is a minor release i'd hate to see a major one

xtqqczze:
@cuviper Would it be possible to first cut a new release after updating to `rand 0.9` (https://github.com/rust-num/num-bigint/pull/322)? It would be great to have a release available with a MSRV of 1.63 before it gets bumped to 1.85.

cuviper:
@xtqqczze Just for MSRV, can't you keep using the current rand support? As I see it, 0.9 would only be useful if you *also* need interoperability with other crates using the same rand.

However, I'm also considering that maybe we should support *both* -- e.g. use `rand09` and `rand010` features and keep our public API in modules by the same name, without re-exporting at the root.

xtqqczze:
> @xtqqczze Just for MSRV, can't you keep using the current rand support? As I see it, 0.9 would only be useful if you _also_ need interoperability with other crates using the same rand.

Yes, releasing with `rand 0.9` would help prevent having multiple `rand` versions in the dependency tree, bloating the size of targets, see https://rust-lang.github.io/rust-clippy/master/index.html#multiple_crate_versions

cuviper:
> However, I'm also considering that maybe we should support _both_

This wasn't too hard to manage -- please take a look!

jgrund:
Hi, anything stopping this from being merged / released?

xtqqczze:
Blocking:

- https://github.com/rust-num/num-bigint/issues/341

PR #339, opened 2026-02-11
Documentation for GCD

mikem8891:
This pull request makes the wording consistent with the wording in the `num-integer` crate.  Note, `BigUint` does not need the statement since it can not be negative.

Related pull request: rust-num/num-integer#75


cuviper:
Thanks!

PR #342, opened 2026-04-12
Update alternatives section in README

xtqqczze:
Update MSRV for alternative crates and use SPDX format for licences.


cuviper:
Thanks!

PR #343, opened 2026-04-21
Remove an unreachable endianness bug in `mac_digit`

cuviper:
This code wrongly created a `[hi, lo]` array to add its final carry digit, when that should be little-endian `[lo, hi]`. However, this was dead code because `mac_with_carry` always shifts `>> BITS`, so the `carry` value **never** has high bits at this point. We'll keep a debug assertion for that, but otherwise it's safe to assume.


cuviper:
AFAICT this bug was introduced in #141, but it was unreachable even then.

PR #344, opened 2026-05-07
clippy: fix manual_clear lint

xtqqczze:
https://rust-lang.github.io/rust-clippy/master/index.html#manual_clear


cuviper:
Thanks, but this is no longer applicable because #307 made the same change.

PR #350, opened 2026-07-06
Improve the implementation of Burnikel-Ziegler division

cuviper:
- Use a block size based on the divisor rather than the dividend
  - This corresponds to the paper's Algorithm 3
- Normalize the divisor just once, avoiding the Knuth shifts as well


cuviper:
I think there's still room to improve, especially avoiding allocations, but so far that's not showing up much in perf profiles. I guess it reduces to the smaller algorithms well enough, and otherwise it depends mostly on multiplication.

PR #352, opened 2026-07-25
Remove unnecessary casts

FranciscoTGouveia:
I believe these casts can be safely removed, as the source type is already the target type.

For reference, these changes were found by an optimizer developed as a research project.

CC @nunoplopes

### DIFF
--- a/src/bigint/bits.rs
+++ b/src/bigint/bits.rs
@@ -474,7 +474,7 @@ pub(super) fn set_negative_bit(x: &mut BigInt, bit: u64, value: bool) {
             // so initially we have `carry_in` = `carry_out` = 1. Furthermore, we
             // stop traversing the digits when there are no more carries.
             let bit_index = (bit / bits_per_digit).to_usize().unwrap();
-            let bit_mask = (1 as BigDigit) << (bit % bits_per_digit);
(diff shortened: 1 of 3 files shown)


src/bigint/bits.rs
```diff
@@ -474,7 +474,7 @@ pub(super) fn set_negative_bit(x: &mut BigInt, bit: u64, value: bool) {
             // so initially we have `carry_in` = `carry_out` = 1. Furthermore, we
             // stop traversing the digits when there are no more carries.
             let bit_index = (bit / bits_per_digit).to_usize().unwrap();
-            let bit_mask = (1 as BigDigit) << (bit % bits_per_digit);
+            let bit_mask = 1 << (bit % bits_per_digit);
```
cuviper: I think there's some value in keeping the explicit types for clarity, but that doesn't have to use `as`. For example, this one could be:


```suggestion
            let bit_mask: BigDigit = 1 << (bit % bits_per_digit);
```

FranciscoTGouveia: Yes, agreed.
I've changed the cast to a type annotation :)

cuviper:
Thanks!

PR #354, opened 2026-08-13
NTT multiplication: rewrite planner and improve comments

byeongkeunahn:
* Rewrote the NTT planner. The planner now considers all possible values of the base naive-multiplication chunk length `g`, which was previously chosen heuristically and then fixed. The timing data was recollected by enumerating all possible `g` values for input lengths up to 300M bits, and the coefficients were then rederived via linear regression. The new fit, which takes `g` into account, achieved `R^2 > 0.99`, and the length-dependent weights turned out to be unnecessary. The `radix-6 + radix-2 → radix-4 + radix-3` plan rewrite also turned out to be slower, so it was removed. The planner now considers `g` in the range `[4, 9]`, since `g = 2` and `g = 3` are never selected by an optimal plan.
* Added and improved comments.

### DIFF
--- a/src/biguint/ntt.rs
+++ b/src/biguint/ntt.rs
@@ -5,16 +5,18 @@ use core::cmp::max;
 use core::iter::zip;
 
 mod arith {
-    // Extended Euclid algorithm:
+    // Extended Euclidean algorithm:
     //   (g, x, y) is a solution to ax + by = g, where g = gcd(a, b)
     const fn egcd(mut a: i128, mut b: i128) -> (i128, i128, i128) {
         assert!(a > 0 && b > 0);
+
+        // treat as a row-major 2x2 matrix
         let mut c = if a > b {
             (a, b) = (b, a);
             [0, 1, 1, 0]
         } else {
             [1, 0, 0, 1]
-        }; // treat as a row-major 2x2 matrix
+        };
         loop {
             if a == 0 {
                 break (b, c[1], c[3]);
@@ -81,9 +83,9 @@ impl<const P: u64> Arith<P> {
         // We only test p^((P-1)/q) != 1 for q in {2, 3, 5} — the prime
         // factors of MAX_NTT_LEN. This suffices because:
         //   - FACTORS_2/3/5 extract all factors of 2, 3, 5 from P-1, so
-        //     R = (P-1)/MAX_NTT_LEN is coprime to MAX_NTT_LEN by construction.
+        //     Q = (P-1)/MAX_NTT_LEN is coprime to MAX_NTT_LEN by construction.
         //   - Any remaining prime factors of P-1 (e.g. 7, 13, 17 for the
-        //     current primes) are absorbed by the p^R exponentiation and
+        //     current primes) are absorbed by the p^Q exponentiation and
         //     do not affect the order of ROOTR.
         //
         // P must be prime for Z/PZ to be a field; this is not checked here.
@@ -100,6 +102,7 @@ impl<const P: u64> Arith<P> {
         }
     };
 
+    // Explicit instantiations to ensure const evaluation at compile time.
     const FACTORS_2: u32 = Self::factors(2);
     const FACTORS_3: u32 = Self::factors(3);
     const FACTORS_5: u32 = Self::factors(5);
@@ -142,7 +145,7 @@ impl<const P: u64> Arith<P> {
         Self::mreduce(lo as u128 | ((hi as u128) << 64))
     }
 
-    // Computes base^exponent mod P with Montgomery reduction
+    // Computes base^exponent mod P with Montgomery reduction.
     const fn mpowmod(mut base: u64, mut exponent: u64) -> u64 {
         let mut cur = Self::R;
         while exponent > 0 {
@@ -157,7 +160,7 @@ impl<const P: u64> Arith<P> {
 
     // Computes c as u128 * mreduce(v) as u128, using d: u64 = mmulmod(P-1, c).
     //
-    // It is caller's responsibility to ensure that d is correct.
+    // It is the caller's responsibility to ensure that d is correct.
     // Note that d can be computed by calling mreducelo(c).
     const fn mmulmod_noreduce(v: u128, c: u64, d: u64) -> u128 {
         let a: u128 = c as u128 * (v >> 64);
@@ -176,12 +179,12 @@ impl<const P: u64> Arith<P> {
         ((m as u128 * P as u128) >> 64) as u64
     }
 
-    // Computes a + b mod P, output range [0, P)
+    // Computes a + b mod P, output range [0, P).
     const fn addmod(a: u64, b: u64) -> u64 {
         Self::submod(a, P.wrapping_sub(b))
     }
 
-    // Computes a + b mod P, output range [0, 2^64)
+    // Computes a + b mod P, output range [0, 2^64).
     const fn addmod64(a: u64, b: u64) -> u64 {
         let (out, overflow) = a.overflowing_add(b);
         if overflow {
@@ -191,7 +194,7 @@ impl<const P: u64> Arith<P> {
         }
     }
 
-    // Computes a + b mod P, selects addmod64 or addmod depending on INV && TWIDDLE
+    // Computes a + b mod P, selects addmod64 or addmod depending on INV && TWIDDLE.
     const fn addmodopt_invtw<const INV: bool, const TWIDDLE: bool>(a: u64, b: u64) -> u64 {
         if INV && TWIDDLE {
             Self::addmod64(a, b)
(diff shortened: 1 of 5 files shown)


src/biguint/ntt.rs
```diff
@@ -212,144 +215,177 @@ impl<const P: u64> Arith<P> {
...
+impl PlanCost {
+    fn new(len: usize, unit_cost: u32) -> Self {
+        if let Ok(unit_cost_usize) = usize::try_from(unit_cost) {
+            if let Some(cost) = len.checked_mul(unit_cost_usize) {
+                return Self::Small(cost);
+            }
+        }
+        Self::Large(len as u128 * unit_cost as u128)
+    }
+}
```
cuviper, line 249: I don't think this `PlanCost` is helpful over just using `u128` unconditionally. The `enum` makes it up to twice as big for the aligned variant tag, and both `new` and `PartialOrd::le` are more complicated as well. See <https://rust.godbolt.org/z/1s7xennTz>.

src/biguint/ntt.rs
```diff
@@ -212,144 +215,177 @@ impl<const P: u64> Arith<P> {
...
+        // For a candidate convolution length `len = g * m`, the estimated cost is
+        // `len * (G_COSTS[g - G_MIN] + sum(radix_counts[i] * RADIX_COSTS[i]))`.
+        // Each radix-r stage processes all `len` elements in `len / r` groups of
+        // r-point butterflies, so the cost of a stage is modeled as linear in
+        // `len`. The constants below represent the relative per-element costs of
+        // the naive base cases and radix butterflies and were calibrated
+        // empirically. `G_COSTS` covers g = G_MIN..=G_MAX, while `RADIX_COSTS`
+        // covers radix-2 through radix-6.
+        const G_COSTS: [u32; G_MAX - G_MIN + 1] = [218, 271, 346, 436, 568, 656];
+        const RADIX_COSTS: [u32; 5] = [504, 785, 754, 1272, 916];
```
cuviper: Please move this directly after `G_MAX`, so we're not interleaving runtime code.

src/biguint/ntt.rs
```diff
@@ -212,144 +215,177 @@ impl<const P: u64> Arith<P> {
...
+        let cnt4 = m2 / 2;
+        m2 -= cnt4 * 2;
+        let [cnt2, cnt3] = [m2, m3];
+
+        [cnt2, cnt3, cnt4, cnt5, cnt6]
+    }
+
+    fn build<const P: u64>(min_len: usize) -> Self {
+        // Allowed values for naive multiplication chunk length (base case).
+        const G_MIN: usize = 4;
```
cuviper, line 298: The PR description states that we start at 4 "since g = 2 and g = 3 are never selected by an optimal plan." This deserves some comment in the code as well, including a brief explanation *why* if possible. Is that just empirical that your exhaustive search never selected 2 or 3?

src/biguint/ntt.rs
```diff
@@ -212,144 +215,177 @@ impl<const P: u64> Arith<P> {
...
+                        None => break,
+                    };
+                    let mut ntt_len = match pow5.checked_mul(pow3) {
+                        Some(ntt_len) => ntt_len,
+                        None => break,
+                    };
+                    if ntt_len / 2 >= ntt_min_len {
+                        // Too much padding will not be computationally optimal
+                        break;
                     }
```
cuviper, line 352: Would it be worth adding an early break in the `m5` loop as well, in case `pow5` alone got too big?

cuviper (changes requested):
> The timing data was recollected by enumerating all possible `g` values for input lengths up to 300M bits, and the coefficients were then rederived via linear regression.

Can you share the code/scripts/etc. that you used for this? If it's manageable, it might even be nice to add that to the repo somewhere.

src/biguint/ntt.rs
byeongkeunahn: That could be worth considering, though the benchmarks didn’t show any benefit in practice.

src/biguint/ntt.rs
byeongkeunahn: Fixed.

src/biguint/ntt.rs
byeongkeunahn: I initially used `u128`, but this was empirically faster—by about 20–30% in `NttPlan::build`. That said, since the NTT itself dominates runtime, I’m fine with using `u128` for simplicity.

I’m not sure yet why `PlanCost` is empirically faster.

src/biguint/ntt.rs
byeongkeunahn: If the NTT radix decomposition contains one of {4, 5, 6}, it can be exchanged with g to improve the cost. If it contains only {2, 3}, those radices can instead be absorbed into g. I added a comment explaining this in the code.

byeongkeunahn:
I have updated the comments and applied the suggestions. I’ll try to share the scripts as well, though it may take me some time to clean them up.

src/biguint/ntt.rs
cuviper: Something weird happened here -- you moved the comment block, but not the actual constants.

src/biguint/ntt.rs
cuviper: That's fine, thanks for trying it.

src/biguint/ntt.rs
cuviper: Do you have a benchmark that you're running on `build` alone?

src/biguint/ntt.rs
byeongkeunahn: Sorry about that—my mistake. Fixed now.

src/biguint/ntt.rs
byeongkeunahn: Yes. I have revised and pushed the benchmark. The timing improvements are around 8%.

----

PlanCost is faster in all 36 cases. Times are in ns/iteration; percentages show the reduction in build time.

| Size | P1 without | P1 PlanCost | Δ | P2 without | P2 PlanCost | Δ | P3 without | P3 PlanCost | Δ |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 2^9  | 742.97 | 674.08 | -9.3% | 799.41 | 731.66 | -8.5% | 795.05 | 730.00 | -8.2% |
| 2^10 | 888.06 | 806.93 | -9.1% | 975.00 | 893.02 | -8.4% | 968.85 | 885.53 | -8.6% |
| 2^11 | 1,000.51 | 910.51 | -9.0% | 1,157.79 | 1,062.60 | -8.2% | 1,162.58 | 1,057.92 | -9.0% |
| 2^12 | 1,131.15 | 1,028.78 | -9.1% | 1,371.47 | 1,256.68 | -8.4% | 1,365.17 | 1,251.66 | -8.3% |
| 2^13 | 1,284.50 | 1,159.30 | -9.7% | 1,607.70 | 1,471.28 | -8.5% | 1,580.22 | 1,448.98 | -8.3% |
| 2^14 | 1,390.41 | 1,278.08 | -8.1% | 1,856.46 | 1,701.93 | -8.3% | 1,788.29 | 1,639.12 | -8.3% |
| 2^15 | 1,534.62 | 1,422.35 | -7.3% | 2,110.90 | 1,936.63 | -8.3% | 2,005.65 | 1,837.98 | -8.4% |
| 2^16 | 1,737.23 | 1,549.55 | -10.8% | 2,416.07 | 2,217.59 | -8.2% | 2,226.95 | 2,045.33 | -8.2% |
| 2^17 | 1,809.40 | 1,656.02 | -8.5% | 2,729.23 | 2,504.28 | -8.2% | 2,459.40 | 2,256.69 | -8.2% |
| 2^18 | 1,974.81 | 1,802.40 | -8.7% | 3,043.80 | 2,796.35 | -8.1% | 2,701.84 | 2,479.86 | -8.2% |
| 2^19 | 2,103.33 | 1,929.82 | -8.2% | 3,354.06 | 3,076.08 | -8.3% | 2,907.68 | 2,678.89 | -7.9% |
| 2^20 | 2,262.12 | 2,081.59 | -8.0% | 3,665.63 | 3,373.46 | -8.0% | 3,133.32 | 2,885.91 | -7.9% |

| Prime | Aggregate reduction | Mean per-case reduction | Range |
|---|---:|---:|---:|
| P1 | 8.73% | 8.82% | 7.3–10.8% |
| P2 | 8.24% | 8.28% | 8.0–8.5% |
| P3 | 8.21% | 8.29% | 7.9–9.0% |
| Overall | 8.36% | 8.46% | 7.3–10.8% |

byeongkeunahn:
Hi, I’ve recollected the timing data and slightly adjusted the coefficients. You can find the benchmark program and the collected data below.

`cargo fmt` seems to behave differently from my installation, so I’ve also adjusted the formatting to match the CI version.

[ntt-bench.zip](https://github.com/user-attachments/files/31141895/ntt-bench.zip)
[timing-data.csv](https://github.com/user-attachments/files/31141920/timing-data.csv)

